PR #15, opened 2019-09-09
Update psycopg2 to fix issue with the docker image

isra17:
See https://github.com/psycopg/psycopg2-wheels/issues/2

I had the same error when running `docker-compose up`.


wengerm50:
I was unable to recreate the issue but thats most likely due to a difference in the architecture of our host machines.

The tests pass and reading through the issue & psycopg's change logs there was definitely issues around the version we have pinned.  I'll go ahead and accept this.

Thanks for your contribution.

PR #19, opened 2019-12-10
WIP: Add a notes section to the annotation page

AstridKery:
This PR makes the following changes:

- Adds a LabelForm component which has:
    -  use a form instead of just buttons
    -  Have a "notes" section that can be optionally filled out when labeling
    - Have optional skip or discard buttons depending on what parent component is using it
    - Since clicking a label doesn't submit the form, a text field lets you know which label will be submitted (the last one you clicked)

- Modifies the annotation page, History table, admin annotation table, and skew tables to use the labelForm.
- Adds label reason to DataLabel and IRRLog, both of which are places where labeled data is stored
- Adds label reason to be displayed in the history table
- A bunch of format fixes from the automatic formatters I'm using. They look nice so I left them in, but I can remove those changes if you'd prefer.


backend/django/core/utils/utils_model.py
```diff
@@ -46,7 +57,10 @@ def cohens_kappa(project):
     labels_seen = set()
     for d in irr_data:
         d_log = IRRLog.objects.filter(data=d, data__project=project)
-        labels = list(set(d_log.values_list('label', flat=True)))
+        # NOTE: Need to figure out why tests fail when this is not there
+        print(list(d_log), len(d_log))
```
AstridKery: So this line has been causing me no end of problems. For some reason, when I get rid of this print statement both cohens kappa tests that expect and answer other than 1 consistently fail. I tried replacing the statement with time.sleep(0.5) thinking that it was some sort of race condition, but that didn't fix it.

PR #20, opened 2019-12-12
Add notes to labeling

AstridKery:
This PR makes the following changes:

- Adds a LabelForm component which has:

   - use a form instead of just buttons
   - Have a "notes" section that can be optionally filled out when labeling
   - Have optional skip or discard buttons depending on what parent component is using it
   - Since clicking a label doesn't submit the form, a text field lets you know which label will be submitted (the last one you clicked)
   - Modifies the annotation page, History table, admin annotation table, and skew tables to use the labelForm.

- Adds label reason to DataLabel and IRRLog, both of which are places where labeled data is stored

- Adds label reason to be displayed in the history table

- A bunch of format fixes from the automatic formatters I'm using. They look nice so I left them in, but I can remove those changes if you'd prefer.


wengerm50:
I would prefer to move the format fixes to their own branch and incorporate the linters/formatters that we've been using on other projects into CI to enforce the change in formatting.

AstridKery:
Yes that makes sense.

wengerm50:
What do you think about radio buttons or a dropdown for the label?  I'm not a huge fan of having the button and text field.

AstridKery:
I think it would be good to have 
- radio buttons if there are < 5 labels
- a dropdown otherwise

Is this the right branch to put that though?

PR #34, opened 2019-12-20
Add notes to labeling

AstridKery:
This branch adds the "notes" section to labels, and allows you to view and edit it in the tables where it's applicable

Another change is that if you have 5+ labels, it renders as a dropdown, and otherwise it renders as a set of radio buttons instead of the previous version which rendered one button per label.


wengerm50:
Can we fix the layout of the labeling component in the following ways:

- [ ] When the label selection uses dropdown, have the label_reason input on the line below
- [ ] Have the skip & submit button on the same line (similar to the discard/submit button on the Skipped page)
- [ ] Add a bit of margin between the label and input box for the label reason

wengerm50:
What do you think about keeping track of why a data was skipped? We'd need to change how we treat the skip button, but it would allow an admin to see why a specific data was skipped.  Maybe we can add this as a separate task in the backlog?

backend/django/core/utils/utils_model.py
```diff
@@ -56,6 +56,9 @@ def cohens_kappa(project):
     labels_seen = set()
     for d in irr_data:
         d_log = IRRLog.objects.filter(data=d, data__project=project)
+        # NOTE: Need to figure out why tests fail when this is not there
+        print(list(d_log), len(d_log))
```
wengerm50: Is this something you still plan on looking into? Or need another set of eyes on?  I'd like to have this figured out instead of leaving as a TODO

backend/django/core/views/api_annotate.py
```diff
@@ -169,6 +169,10 @@ def annotate_data(request, data_pk):
     label = Label.objects.get(pk=request.data["labelID"])
     labeling_time = request.data["labeling_time"]
 
+    label_reason = ""
+    if "labelReason" in request.data:
+        label_reason = request.data["labelReason"]
```
wengerm50: This can be shortened to

`label_reason = request.data.get("labelReason", "")`

wengerm50: This actually appears in a few spots further down so you can change in all those places please

backend/django/core/utils/utils_model.py
AstridKery: I put the comment there in case you had some ideas. I'd like to fix this before merging, but I'm not sure what could be the cause of it.

AstridKery:
> What do you think about keeping track of why a data was skipped? We'd need to change how we treat the skip button, but it would allow an admin to see why a specific data was skipped. Maybe we can add this as a separate task in the backlog?

So the `label_reason` field has been added to both the "DataLabel" and the "IRRLog" tables as properties of labeled data. We don't currently have a table for skipped data (it just gets put in a queue) so as things are now there isn't a straightforward way to add reasons to the skipped data. 

We could theoretically create a new table "data annotations" which stores notes about data objects along with the context of the notes (irr, skipped, or labeled). That way we could also save the notes from when different people label IRR which could be helpful for the admin for resolving conflict.

What do you think?

backend/django/core/utils/utils_model.py
wengerm50: I've addressed this in bb58620

PR #37, opened 2020-01-08
Ctp fixes

AstridKery:
This branch addresses the following issues found with SMART after the last update:

-  The date parsing bogs down with not-tiny datasets. Needed to use the version of the dateparser library's parser which is specifically made for large datasets
- The URL checking right now throws an error if any URL doesn’t work without any information. This has been changed to tell the user exactly which url was invalid.
- The fix skew page table takes a while to populate for large datasets. Added a note so the users know they need to wait
- The label reason and metadata should be downloaded with the labels
- The past reason should show up in the history table reason input
- The project update page needs to have the data directions updated
- Fix the sample csv to have correct titles
- Make the input restrictive on the names allowed


frontend/src/components/LabelForm/index.jsx
```diff
@@ -234,7 +234,7 @@ class LabelForm extends React.Component {
                         (Optional) Reason for Label:
                         <FormControl
                             componentClass="textarea"
-                            placeholder="Type reason here"
+                            placeholder={(this.state.label_reason === "") ? "Type reason here" : this.state.label_reason}
```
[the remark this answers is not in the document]
AstridKery: So that makes sense for the history table, but it gets really annoying when you're trying to input a new reason in the annotation page. If you erase it it just pops back in.

frontend/src/components/Skew/index.jsx
```diff
@@ -33,8 +33,40 @@ class Skew extends React.Component {
         this.props.getLabelCounts();
     }
 
+    renderTable(){
+        const { unlabeled_data, skewLabel, labels } = this.props;
+        if (unlabeled_data.length == 0) {
+            return <b>Loading unlabeled data. This may take up to a minute for large datasets..</b>;
```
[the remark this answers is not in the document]
AstridKery: So it is trying to load all of the data. I could see pagination being helpful, though I don't know how simple it would be to use the search part of the table with it. I suppose the search functionality could be moved to a separate thing? It seems pretty important for someone to search for content that is related to a particular label.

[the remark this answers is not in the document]
AstridKery: I added backend pagination, sorting, and filtering for the skew page. I'm guessing this is going to need to be added for the History page as well (since it will bog down once the number of labeled articles gets large).

frontend/src/components/LabelForm/index.jsx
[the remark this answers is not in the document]
wengerm50: Yea I was commenting with respect to the history component.  If its too much of a hassle to differentiate which parent component the label form is in then I'm fine leaving as is.

frontend/src/components/Skew/index.jsx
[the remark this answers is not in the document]
wengerm50: Agreed. Do you want to add that as part of this mr or just add it to the backlog?

backend/django/test/api/test_api_annotate.py
```diff
@@ -505,17 +500,17 @@ def test_unlabeled_table(
         {"labelID": test_labels[0].pk, "labeling_time": 1},
     )
     response = admin_client.get(
-        "/api/data_unlabeled_table/" + str(project.pk) + "/"
+        "/api/data_unlabeled_table/?project=" + str(project.pk)
```
wengerm50: I'd prefer to start migrating to using f-strings.  It might be odd at the start since all of the old code doesn't use them, but if you make a change that can use f-strings then go ahead

frontend/src/actions/skew.js
```diff
@@ -7,13 +7,35 @@ import { getHistory } from './history';
...
+        if (queryObj.sorted[0].desc) {
+            queryURLObj.append("ordering", "text");
+        } else {
+            queryURLObj.append("ordering", "-text");
+        }
+
+    }
+
+
+    let apiURL = `/api/data_unlabeled_table/?${queryURLObj.toString()}`;
```
wengerm50: I'm not sure if you are able to add a debounce to the search filter for this or not.  I think its a nice-to-have but not necessary if time is an issue.

AstridKery: I don't see an obvious way to do this. I think this just uses onChange for an input so there might not be a concept of "finished typing".

AstridKery: So I added a change in 1ff2926 so that the search only happens when you hit enter. I think this should address the main issue. Thoughts?

frontend/src/components/Skew/index.jsx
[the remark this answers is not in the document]
AstridKery: I'll add it to the backlog. I'd rather get these changes nailed down first.

backend/django/test/api/test_api_annotate.py
AstridKery: So... I realize you were not intending me to replace all the string concats with f-strings.... but once I started doing it I couldn't figure out a good stopping point. So yeah. Most of them are now f-strings.

frontend/src/actions/skew.js
wengerm50: I think I'd prefer it the original way. Since there is no submit button its not clear that pressing enter will start the search.  Also for existing users it directly changes a feature that they are already used to.

backend/django/test/api/test_api_annotate.py
wengerm50: No complaints from me.  If we've already gone this far then might as well go all the way.  There are a few more places that could be replaced.  Use the follow search-all regex

`".*" \+`

AstridKery: I think I got all of them on the backend now.

PR #41, opened 2020-01-24
Refactors Prod Settings

wengerm50:
This is a WIP refactor of the prod docker compose settings. This provides
a very simple & very NOT SECURE method of deploying the application using
nginx & gunicorn.


wengerm50:
Another issue found. When you rebuild in production the old dist files are in the static_files volume. so when you rebuild the frontend and start the compose process, the volume mounts over the new dist files. current solution is to drop the static volume, rebuild, then collectstatic again

envs/prod/rebuild.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# WARNING: This is a convience script made for testing & debugging
+# the production settings.  Do NOT run this in actual produciton
+# or all of your data will be deleted!!!
+docker-compose down
+docker volume rm vol_smart_pgdata_prod vol_smart_data_prod
```
wengerm50: want to remove static volume here

wengerm50: Done

wengerm50:
One more issue.  Since codebooks are saved in a static directory, they are not immediately available and `collect_static` has to be re-ran.  codebooks are not actually static files but media files so they need to be treated differently.

wengerm50:
Additions:

* Handle static files differently (build at compose-up time not at build time) so that you do not have to delete the static-volume
* Treat codebooks as media so they are available w/o collecting-static
* Increase the timeouts gunicorn & nginx timeouts
* Moves the init_redis into a management command to prevent duplicate data entries in the redis queue

PR #43, opened 2020-01-30
Exclude edits

AstridKery:
This MR does a bunch of smaller things, mostly revolving around the exclude table and data downloads

- Changes to RecycleBin/Excluded items 
   - Freeze all past labels for data when it’s excluded, and have the restoring function put the post back in the admin table with all its past labels and IRR data intact.
   - Make it so excluding data does not drop DataLabels or IRRLogs anymore
   - Ignore excluded data for all the places where DataLabel and IRRLog objects are pulled
      - History table
      - Admin table
      - Assign_data
      - Skew_table
      - Label_counts and tables on admin page
      - Data downloads
      - Irr page charts
      - Model running
      - Handling IRR everywhere
   - Make it so restoring data puts the data back in the admin DataQueue
   - Provide a “reason” field when excluding data, which would be shown in the table

- Changes to Downloading things
   - Add a button for an admin to download excluded data, including the exclusion reason and explicit indicator (possibly past labels, though this could get complicated if the data is IRR and has multiple labels)
   - Add a button for admin to download IRR data. Should include current DataLabels for IRR data and IRRLog entries
   - Make it so the downloaded data doesn’t include IRR (this was a bug that had to be fixed. Before this change, if someone downloaded the labeled data, it would include un-finalized IRR labels)

- Changes to The admin charts
   - Make it so the label counts and label speed metrics DO include IRRLog labels 
   - Add clarifying text to the titles of charts to make it clear when they are showing ALL labels versus FINAL labels
   - Add a chart which gives the number of finalized labels total and for each label


backend/django/core/models.py
```diff
@@ -235,6 +247,7 @@ class TrainingSet(models.Model):
 
 class RecycleBin(models.Model):
     data = models.ForeignKey("Data")
+    exclude_reason = models.TextField(null=True, default="")
     timestamp = models.DateTimeField(default=timezone.now)
```
wengerm50: This comment isnt related to this line necessarily, but to our models in general.  I think it would be helpful to create custom queryset managers for out models.  That way it is easier to access commonly queries things without having to remember which combination of boolean fields need to be true or false or w/e

wengerm50: https://docs.djangoproject.com/en/3.0/topics/db/managers/#creating-a-manager-with-queryset-methods

We also use this pretty extensively on warrants so if you need examples you can check there or ask Peter/myself

AstridKery: This is awesome! I was really disliking copying things into filters everywhere because it felt so easy to make a mistake.

AstridKery:
FYI: I finally chased down one reason why the "duplicate label" issue was happening. The card deck was being pulled whenever the component mounted but if you just changed tabs to history and back the cards state was still there. Thus get_card_deck would be called which would return the same cards it gave last time which were then appended to the existing deck. 

A simple if statement in the mounting function solved the problem.

backend/django/core/utils/util.py
```diff
@@ -555,7 +558,7 @@ def get_excluded_data(project):
             temp.update(MetaDataSerializer(metadata).data)
 
         data_labels = DataLabel.objects.filter(data=d.data)
-        for i in range(data_labels.count()):
+        for i, label in enumerate(data_labels):
             temp[f"label_{i}"] = data_labels[i].label.name
             temp[f"label_{i}_reason"] = data_labels[i].label_reason
```
wengerm50: since you used enumerate you now have access to the item that is in the data_labels list at index i.  So instead of `data_labels[i].label.name` you should just be able to use `label.name`

backend/django/core/utils/utils_model.py
```diff
@@ -124,9 +114,7 @@ def fleiss_kappa(project):
     data_label_dict = []
     num_data = 0
     for d in irr_data:
-        d_data_log = IRRLog.objects.filter(
-            data=d, data__project=project, data__recyclebin__isnull=True
-        )
+        d_data_log = IRRLog.objects.filter(data=d, data__project=project)
```
wengerm50: should this be `IRRLog.objects.unexcluded()`?

backend/django/core/utils/utils_queue.py
```diff
@@ -266,12 +263,15 @@ def pop_first_nonempty_queue(project, profile=None, type="normal"):
...
+            skipped_data = (
+                IRRLog.objects.skipped()
+                .filter(profile=profile)
+                .values_list("data", flat=True)
+            )
 
             assigned_unlabeled = (
-                DataQueue.objects.filter(queue=queue, data__recyclebin__isnull=True)
+                DataQueue.objects.filter(queue=queue)
+                .filter(data__recyclebin__isnull=True)
```
wengerm50: should this be DataQueue.objects.unexcluded()?

backend/django/core/utils/utils_queue.py
```diff
@@ -376,27 +376,18 @@ def handle_empty_queue(profile, project):
...
     irr_queue = Queue.objects.get(project=project, type="irr")
 
-    queue_count = DataQueue.objects.filter(
-        queue=queue, data__recyclebin__isnull=True
-    ).count()
-    irr_count = DataQueue.objects.filter(
-        queue=irr_queue, data__recyclebin__isnull=True
-    ).count()
+    queue_count = DataQueue.objects.filter(queue=queue).count()
+    irr_count = DataQueue.objects.filter(queue=irr_queue).count()
```
wengerm50: Should these two be `DataQueue.objects.unexcluded`?

wengerm50: Ah we dont have a DataQueue manager yet... In that case do they need to still have `data__recyclebin__isnull=True`?

backend/django/core/utils/utils_queue.py
```diff
@@ -376,27 +376,18 @@ def handle_empty_queue(profile, project):
...
     assigned_toOthers_count = (
-        AssignedData.objects.filter(queue=queue, data__recyclebin__isnull=True)
-        .exclude(profile=profile)
-        .count()
+        AssignedData.objects.filter(queue=queue).exclude(profile=profile).count()
```
wengerm50: Same question as above (AssignedData.objects.excluded() or keep data__recylebin__isnull)

PR #51, opened 2020-02-28
Add links to any urls that appear in annotation cards AND massively update docker/node, package.json and yarn

rjurney:
This uses the package linkifyjs and linkifyjs/react to automatically create http links for any urls present in an annotation card. This enables users to refer to outside resources when labeling. 

In order to achieve that and to make this deployable I had to upgrade the Dockerfile to use `node:latest` as the parent and also upgrade all the yarn/npm packages to later versions. Or I couldn't deploy the thing.


rjurney:
I'm trying to figure out why Docker does not build... it builds locally. 

I wonder if it isn't using the dev setup but some other environment setup that is not in the code for the project?

PR #56, opened 2020-04-23
fix: Filtering a sliced query in fleiss_kappa

mrshu:
* The `fleiss_kappa` function retrieves objects from the IRRLog in order
  to compute the Fleiss Kappa. If the number of retrieved objects is
  larger than the number of labelers (n), the retrieved objects are
  sliced to be at most n in size.

  This, however, leads to the "Cannot filter a query once a slice has
  been taken" error and a 500 response.

* This commit fixes the situation by filtering the retrieved IRRLog
  objects directly in Python instead of doing it on the ORM level.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/backend/django/core/utils/utils_model.py
+++ b/backend/django/core/utils/utils_model.py
@@ -119,7 +119,10 @@ def fleiss_kappa(project):
                 l_label = "skip"
             else:
                 l_label = str(l)
-            label_count_dict[l_label] = len(d_data_log.filter(label__name=l))
+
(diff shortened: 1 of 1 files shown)


mrshu:
@rchew Thanks a ton for putting SMART together -- it's been of quite some help!

I ran into this while using it with more than 10 users. This fix has covered an error we were running into in the Admin's IRR section.

Please do let me know if there are any other things that you believe should change for this PR to be merged in.

Thanks!

PR #101, opened 2021-07-23
add searchable dropdown for labels

dsteedRTI:
This keeps the buttons for labels the same if there are less than 5 labels; however, if there are many labels a searchable dropdown is generated with all of the label options


frontend/src/components/DataCard/index.jsx
```diff
@@ -32,22 +35,43 @@ class DataCard extends React.Component {
...
-                                    {opt["name"]}
-                                </Button>
-                            ))}
+                                    options={labelsOptions}
+                                    placeholder="Select label..."
+                                    searchBy="name"
+                                    sortBy="name"
+                                    style={{ minWidth: '200px' }}
+                                />
+                            :
```
AstridKery, line 57: So when I run this, I get a linting error for this line. Did the pre-commit linting run properly when you committed this file? You can see the error if you run this `docker-compose -f envs/dev/docker-compose.yml run --rm smart_frontend yarn lint`

frontend/package.json
```diff
@@ -36,45 +36,39 @@
...
-    "popper.js":"1.14.3",
+    "popper.js": "1.14.3",
     "prop-types": "15.7.2",
-
     "react": "16.14.0",
+    "react-bootstrap": "1.5.2",
     "react-dom": "16.14.0",
-    "react-test-renderer": "16.14.0",
-
+    "react-dropdown-select": "^4.7.4",
```
AstridKery, line 64: So I'm not totally against the use of `^` here, but I'm also not quite sure if we should use `^`, `~`, or nothing. For differences between the two see: https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json

Right now, all our other packages just specify the version, so for consistency we might want to keep it specific. On the other hand, it might be useful to get patches. @peridoteagle what do you think?

AstridKery:
Thanks @dsteedRTI ! The label dropdown looks great in the annotation page. However, while we're at it we should probably make this change for the other places where the set of labels are currently rendered individually. That is:

* the fix skew table
* the history table
* The admin irr/skipped table

frontend/src/components/DataCard/index.jsx
dsteedRTI: Good catch. I'm not sure why mine did not run, but I think that file should be fixed now

dsteedRTI:
> Thanks @dsteedRTI ! The label dropdown looks great in the annotation page. However, while we're at it we should probably make this change for the other places where the set of labels are currently rendered individually. That is:
> 
> * the fix skew table
> * the history table
> * The admin irr/skipped table

Just added the dropdown search to all 3. 
Some more complicated styling was needed for the admin table due to the table's CSS constrains I think, but it works now.

frontend/package.json
AstridKery: After some discussion I think this is fine. We might want to make a branch to update all of the dependencies to use ^, but that's outside the scope of this PR

AstridKery:
Looks good to me. I'll merge it in. Thanks!

PR #102, opened 2021-07-27
add upload option for labels with .csv files

dsteedRTI:
this allows users to upload a .csv files with labels. after we discussed yesterday, I just used js to populate the form with the entries from the csv

### DIFF
--- a/backend/django/core/templates/projects/create/create_wizard_labels.html
+++ b/backend/django/core/templates/projects/create/create_wizard_labels.html
@@ -11,6 +11,12 @@
 {% endblock %}
 
 {% block content %}
+<div class="overlay">
+  <div class="overlay_contents">
+      <div id="second_progress"></div>
+      <span id="loading_text"></span>
+  </div>
+</div>
 <div class="error-messages">{{ wizard.non_form_errors }}</div>
 <div class="row">
   <div class="col-md-8 col-md-offset-2">
@@ -31,7 +37,13 @@
             <li class="list-group-item">If you plan on uploading a data file that contains labels, the label categories in the file must match those provided on this page.</li>
             <li class="list-group-item">You cannot update the labels for a project after the project is created.</li>
           </ul>
-          <table class="table table-striped">
+          <div class="form-group">
+            <label class="control-label" for="{{ wizard.form.data.id_for_label }}">You may also upload a .csv file containing labels and their descriptions (label and description columns required):</label>
+            <label class="control-label" for="{{ wizard.form.data.id_for_label }}">{{ wizard.form.data.label }}</label>
+            <input class="form-control" id="{{ wizard.form.data.id_for_label }}" maxlength="30" name="{{ wizard.form.data.html_name }}" type="file" placeholder="{{ form.data.label }}" onChange="handleUpload(event)" onclick="this.value = null;" />
+          </div>
+          <br />
+          <table class="table table-striped" id="form-table">
             {{ wizard.form.management_form }}
             <div class="error-messages">{{ wizard.form.non_form_errors }}</div>
             {% for form in wizard.form %}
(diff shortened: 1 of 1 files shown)


backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -73,15 +85,60 @@ <h3>Instructions</h3>
...
+              }
+            });
+            if($('.formset_row_label').length > data.length -1) {
+              $('.formset_row_label').slice(data.length - 1).remove();
+            }
+            $('.overlay').hide();
+        }
+      } catch (error) {
+        $('.error-messages').empty();
+        $('.error-messages').eq(1).append('<ul class="error-list"><li>There was an error processing your label file. Please try again.</li></ul>');
```
AstridKery, line 127: Not a big problem, but I noticed if I tried to upload a bad file, and then tried again with a correct file, the labels would fill but the error message would remain

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -31,7 +37,13 @@ <h3>Instructions</h3>
             <li class="list-group-item">If you plan on uploading a data file that contains labels, the label categories in the file must match those provided on this page.</li>
             <li class="list-group-item">You cannot update the labels for a project after the project is created.</li>
           </ul>
-          <table class="table table-striped">
+          <div class="form-group">
+            <label class="control-label" for="{{ wizard.form.data.id_for_label }}">You may also upload a .csv file containing labels and their descriptions (label and description columns required):</label>
```
AstridKery, line 41: Note: when this happens any existing labels in the label fields are overwritten by the file. This isn't wrong, but we may want to note that here.

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -73,15 +85,60 @@ <h3>Instructions</h3>
...
+            }
+            $('.overlay').hide();
+        }
+      } catch (error) {
+        $('.error-messages').empty();
+        $('.error-messages').eq(1).append('<ul class="error-list"><li>There was an error processing your label file. Please try again.</li></ul>');
+      }
+    } else {
+      $('.error-messages').empty();
+      $('.error-messages').eq(1).append('<ul class="error-list"><li>There was an error processing your label file. Please make sure your file is a .csv and has a Label and Description column.</li></ul>');
```
AstridKery, line 131: So since we have it available for the data upload page, we should probably have a sample label file for people to download from the label upload page.

Also just noticed that the sample data file for the data page is using the old data source...-_-,

AstridKery:
@dsteedRTI great solution to a complicated ask! I have a few discussions, but in general this works really well.

backend/django/core/templates/projects/create/create_wizard_labels.html
dsteedRTI: done

backend/django/core/templates/projects/create/create_wizard_labels.html
dsteedRTI: added

backend/django/core/templates/projects/create/create_wizard_labels.html
dsteedRTI: I went ahead and made sample label .csv and updated the example data to have the new r/awww data

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -73,15 +88,59 @@ <h3>Instructions</h3>
...
+      try {
+        reader.readAsText(event.target.files[0]);
+        reader.onload = function(event){
+          var csv = event.target.result;
+          var data = $.csv.toArrays(csv);
+            data.forEach((label, index) => {
+              if(index !== 0) {
+                if($('.formset_row_label').eq(index - 1).length > 0) {
+                  $('.formset_row_label').eq(index - 1).css('display', '');
+                  $('.formset_row_label').eq(index - 1).find('input[type=text]').eq(0).val(label[0]);
```
AstridKery, line 109: Noticed something new with this. It isn't strictly *wrong*, but we might want to add a check before this line making sure the fields provided are label and description. 

Basically, I found out if you feed this data input one of the sample data files, it takes the first column "Text" as the list of labels with "Label" as the descriptions, thus making one label for each data point.

Thoughts?

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -73,15 +88,59 @@ <h3>Instructions</h3>
...
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-csv/0.71/jquery.csv-0.71.min.js"></script>
 <script type="text/javascript">
-$('.formset_row_label').formset({
-  addText: 'add label',
-  deleteText: 'remove label',
-  prefix: 'label_set',
-  added: function() {
-    // Reset drop down and errors from both columns when adding new rows
-    $('tr.formset_row_label').last().children(':first').find('ul.errorlist li').empty();
+  function handleUpload(event) {
```
AstridKery, line 93: Found a new issue with adding a lot of labels which we'll need to resolve before merging in. I tried to add a file with thousands of labels. It added them correctly to the table, but upon submitting I got this error 

Screen Shot 2021-08-02 at 5 10 52 PM

I suspect this is an error which could be solved by updating a constant in `backend/django/smart/settings.py`. See this post for reference https://stackoverflow.com/questions/47585583/the-number-of-get-post-parameters-exceeded-settings-data-upload-max-number-field

dsteedRTI:
the last commit ^ solved the error with handling thousands of labels, that should work now

AstridKery (approved):
Looks good!

PR #103, opened 2021-07-28
improve label display for admin distribution chart

dsteedRTI:
This affects the distribution chart on the admin page:

- Label legend is now on the bottom of the chart
- Both distribution and time chart span the entire width
- The top N (currently = 5) labels are displayed in the legend, along with a final "other labels" tab

### DIFF
--- a/backend/django/core/views/api_admin.py
+++ b/backend/django/core/views/api_admin.py
@@ -40,9 +40,19 @@ def label_distribution(request, project_pk):
     users.append(project.creator)
     users.extend([perm.profile for perm in project.projectpermissions_set.all()])
 
+    def sortCount(label):
+        label_count = 0
+        for u in users:
+            label_count += DataLabel.objects.filter(profile=u, label=label).count()
+        return label_count
+
+    # Sort labels by the count
+    labels.sort(key=sortCount, reverse=True)
+
     dataset = []
     all_counts = []
-    for label in labels:
+    # Get first labels
+    for label in labels[0:5]:
         temp_values = []
         for u in users:
             label_count = DataLabel.objects.filter(profile=u, label=label).count()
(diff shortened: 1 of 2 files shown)


backend/django/core/views/api_admin.py
```diff
@@ -40,18 +40,34 @@ def label_distribution(request, project_pk):
...
     users.extend([perm.profile for perm in project.projectpermissions_set.all()])
 
+    def sortCount(label):
+        label_count = 0
+        for u in users:
+            label_count += DataLabel.objects.filter(profile=u, label=label).count()
+        return label_count
+
+    # Sort labels by the count
+    labels.sort(key=sortCount, reverse=True)
```
AstridKery, line 50: I think the goal of the sortCount function and this line is to sort the label list by the total number of datalabels right? I think you can simplify this to one line

`labels.sort(key=lambda label: DataLabel.objects.filter(label=label).count(), reverse=True)`

backend/django/core/views/api_admin.py
```diff
@@ -40,18 +40,34 @@ def label_distribution(request, project_pk):
...
+
     dataset = []
     all_counts = []
```
AstridKery, line 53: It doesn't look like all_counts is used for anything anymore. I think we can just remove it. 

We might want to think about what should be returned if there is no data. Right now if we look at the admin chart for the "no labels" project, we see this:

Screen Shot 2021-07-29 at 2 03 22 PM

I don't hate this, but we could alternatively show a message like we do in the IRR chart.

AstridKery:
Looks good @dsteedRTI ! I have a few comments, and we should probably merge npsas_main into this branch to get it up to date.

backend/django/core/views/api_admin.py
dsteedRTI: I modified the backend some to only return data when counts > 0, and that solved the issue with the chart rendering but having no data. It now looks like the IRR chart when there is no data

PR #106, opened 2021-08-09
Add metadata option to the data

AstridKery:
This branch adds the ability for users to upload datasets with metadata. Currently all extra columns in the tables will be automatically read in as metadata.

The changes in this MR:

* add a sample data file with metadata
* add metadata field and metadata tables to the database to log all metadata values
* update the data ingest for creating and updating projects so metadata fields are accepted
* update all places on the annotate page where data is being passed and displayed so metadata fields are displayed


backend/django/core/views/api_admin.py
```diff
@@ -30,7 +30,7 @@ def label_distribution(request, project_pk):
 
     Args:
         request: The POST request
-        project_pk: Primary key of the project
+        project_pk: Primary key o the project
```
dsteedRTI, line 33: I think there might be a typo here? Was this accidental?

dsteedRTI (approved):
Tested and appears to be working! I can tell this was a lot of work but great job!

Code is very readable and you kept everything consistent across the files in frontend/backend.

backend/django/core/views/api_admin.py
AstridKery: I was feeling old fashioned :). No it was a typo I'll fix it.

PR #107, opened 2021-08-09
improve irr chart visibility by selection union of labels to be displ…

dsteedRTI:
This updates the readability of the irr chart by trimming down the labels on the X and Y axis to the union of labels used by the two users selected

### DIFF
--- a/backend/django/core/views/api_admin.py
+++ b/backend/django/core/views/api_admin.py
@@ -312,10 +312,21 @@ def heat_map_data(request, project_pk):
     project = Project.objects.get(pk=project_pk)
 
     heatmap_data = irr_heatmap_data(project)
-    labels = list(
-        Label.objects.all().filter(project=project).values_list("name", flat=True)
-    )
-    labels.append("Skip")
+
+    allLabels = [label for label in project.labels.all()]
+
+    users = []
+    users.append(project.creator)
+    users.extend([perm.profile for perm in project.projectpermissions_set.all()])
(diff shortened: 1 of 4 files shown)


backend/django/core/views/api_admin.py
```diff
@@ -312,10 +312,21 @@ def heat_map_data(request, project_pk):
     project = Project.objects.get(pk=project_pk)
 
     heatmap_data = irr_heatmap_data(project)
-    labels = list(
-        Label.objects.all().filter(project=project).values_list("name", flat=True)
-    )
-    labels.append("Skip")
```
AstridKery, line 318: So I believe the point of this line was that for some IRR, one user might label it while the other skips it. This counts as a disagreement.

frontend/src/utils/admin_irr.js
```diff
@@ -17,13 +17,17 @@ function drawHeatmap(response) {
     let all_zero = true;
     //don't draw the heatmap if there is no data
     let most_data = 0;
-    for (let i = 0; i < data.length; i++) {
```
AstridKery, line 20: Screen Shot 2021-08-18 at 10 16 42 AM

So I'm running into some rendering problems with this chart. Have you seen this before?

AstridKery:
Hey @dsteedRTI ! Just realized this must have fallen through the cracks. Definitely not a priority but we should probably resolve this before the end of the month.

PR #110, opened 2021-08-17
add desription to labels, top 5 label dropdown, and outstanding items #

dsteedRTI:
this adds:
- descriptions to the label preview in the dropdown menu
- top 5 label dropdown on top right of card:
Note: This only uses the first five labels for now, but Emily said this will be based off a model soon
- number of outstanding items below cards on the Annotate tab


AstridKery:
Screen Shot 2021-08-18 at 10 16 42 AM
I'm running into some rendering issues with the heatmap rendering on top of the labels

frontend/src/components/DataCard/index.jsx
```diff
@@ -110,6 +110,29 @@ class DataCard extends React.Component {
                                 </Button>
                             </OverlayTrigger>
                         </ButtonToolbar>
+                        {labels.length > 5 && (
+                            <div className="suggestions dropdown">
+                                <button type="button" className="btn btn-info dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                                    Suggested Labels
```
AstridKery, line 116: So for this, I was wondering if maybe we should just have a table instead of a dropdown? That way the user could see the top 5 suggestions without clicking on the dropdown, and would then just enter their choice in the regular text box. What do you think?

frontend/src/components/DataCard/index.jsx
```diff
@@ -110,6 +110,29 @@ class DataCard extends React.Component {
...
+                                </button>
+                                <ul className="dropdown-menu dropdown-menu-right">
+                                    {labelsOptions.slice(0, 5).map(opt => (
```
AstridKery, line 119: So it looks like this will always return the first 5 labels alphabetically. I know that we don't really have a model here, but I think we can start with something a bit simpler that's nevertheless useful.

I think a good starting point would be to make an api call that finds the label+descriptions with the most overlap with the text. 

There are a number of ways to do this. I think an easy one to start would be lemmatizing the text and getting the Jaccard index https://en.wikipedia.org/wiki/Jaccard_index similarity between the text and each possible label+description pair. We could do this upon data upload so the calculations only have to happen once.

Thoughts?

frontend/src/components/UnlabeledDataStatus/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+        this.props.getUnlabeled();
+    }
+
+    render() {
+        const { unlabeled_data } = this.props;
+
+        console.log(this.props);
+        return (
+            <p className="unlabeled-status">
+                Unlabeled Outstanding: {unlabeled_data.length}
```
AstridKery, line 15: Screen Shot 2021-08-18 at 1 49 38 PM

So I ran into this situation where I made a ton of labels and accidentally forgot to set the batch size to something reasonable. You'll notice that the unlabeled count is 0, probably because everything is assigned. I think we might need to split this count into a few parts.

1. Unlabeled and Unassigned
2. Partially labeled IRR
3. Waiting for Admin Review (IRR or Skipped)
4. With final label

Might want to move this to the details page though. Thoughts?

AstridKery:
Hi @dsteedRTI ! Thanks for jumping into this. I have some discussions about each of these on how we could further refine things, so I'd appreciate if you could look through and let me know your thoughts. I'm happy to jump on a call to discuss further. Thanks again!

frontend/src/components/DataCard/index.jsx
dsteedRTI: I can put a table in! I had played around with it some, and when you have both the label and the description I felt like the card felt a little cluttered so I had switched to a dropdown to save space. But if you want a regular table, then I can move it to that!

frontend/src/components/UnlabeledDataStatus/index.jsx
dsteedRTI: Ok that sounds like a good next step! I can get working on adding all of those

frontend/src/components/Smart/index.jsx
```diff
@@ -8,9 +8,11 @@ import AdminTableContainer from '../../containers/adminTable_container';
 import RecycleBinContainer from '../../containers/recycleBin_container';
 import CodebookLabelMenuContainer from '../../containers/codebookLabelMenu_container';
 import SmartProgressBarContainer from '../../containers/smartProgressBar_container';
+import UnlabeledDataStatusContainer from '../../containers/unlabeled_data_status_container';
 
-const ADMIN = window.ADMIN;
 
+const ADMIN = window.ADMIN;
+const PROJECT_ID = window.PROJECT_PK;
```
AstridKery, line 15: The linter finds the following error with this line:

```
/code/src/components/Smart/index.jsx
  15:7  warning  'PROJECT_ID' is assigned a value but never used  no-unused-vars
```

To recreate the error, run the linting command `docker-compose -f envs/dev/docker-compose.yml run --rm smart_frontend yarn lint`

backend/django/core/views/api_annotate.py
```diff
@@ -436,9 +437,14 @@ def data_unlabeled_table(request, project_pk):
     data = []
     for d in unlabeled_data:
         serialized_data = DataSerializer(d, many=False).data
+
```
AstridKery, line 440: I get the following error when clicking on the "annotate" page in any of the three sample projects 

```
smart_backend    | Internal Server Error: /api/data_unlabeled_table/3/
smart_backend    | Traceback (most recent call last):
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
smart_backend    |     response = get_response(request)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
smart_backend    |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
smart_backend    |     return view_func(*args, **kwargs)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
smart_backend    |     return self.dispatch(request, *args, **kwargs)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
smart_backend    |     response = self.handle_exception(exc)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
smart_backend    |     self.raise_uncaught_exception(exc)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
smart_backend    |     raise exc
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
smart_backend    |     response = handler(request, *args, **kwargs)
smart_backend    |   File "/usr/local/lib/python3.8/site-packages/rest_framework/decorators.py", line 50, in handler
smart_backend    |     return func(*args, **kwargs)
smart_backend    |   File "/code/core/views/api_annotate.py", line 443, in data_unlabeled_table
smart_backend    |     sorted_similarity_pairs.append(json.loads(pair))
smart_backend    |   File "/usr/local/lib/python3.8/json/__init__.py", line 357, in loads
smart_backend    |     return _default_decoder.decode(s)
smart_backend    |   File "/usr/local/lib/python3.8/json/decoder.py", line 337, in decode
smart_backend    |     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
smart_backend    |   File "/usr/local/lib/python3.8/json/decoder.py", line 355, in raw_decode
smart_backend    |     raise JSONDecodeError("Expecting value", s, err.value) from None
smart_backend    | json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

backend/django/core/utils/util.py
```diff
@@ -148,6 +150,8 @@ def create_data_from_csv(df, project):
             ],
         )
 
+    create_label_similarity_results(df, project)
```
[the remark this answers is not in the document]
dsteedRTI: done

frontend/src/components/DataCard/index.jsx
```diff
@@ -39,26 +39,37 @@ class DataCard extends React.Component {
...
+                                ))}
+                            </div>
+                        )}
+                        <ButtonToolbar className="btn-toolbar">
                             {labels.length > 5 ? (
                                 <Select
                                     className="align-items-center flex py-1 px-2"
                                     dropdownHandle={false}
-                                    labelField="name"
+                                    labelField="dropdownLabel"
```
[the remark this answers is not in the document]
dsteedRTI: done

backend/django/core/views/api_admin.py
```diff
@@ -323,3 +325,44 @@ def heat_map_data(request, project_pk):
...
+    Args:
+        request: The GET request
+        project_pk: Primary key of the project
+    Returns:
+        data: a list of data information
+    """
+
+    project = Project.objects.get(pk=project_pk)
+
+    total_data_objs = list(project.data_set.filter(datalabel__isnull=True))
```
AstridKery, line 344: So the total here is actually going to be too low, since you're filtering out the labeled data. 

The "other-labeled" value is "total" - "final" - "unlabeled", so "total" should be just the length of `project.data_set`

To test this, check out the three person default project. It's got a negative number right now for "other"

backend/django/core/utils/util.py
```diff
@@ -205,6 +209,27 @@ def create_metadata_objects(df, project):
         MetaData.objects.bulk_create(metadata_objects)
 
 
+def create_label_similarity_results(project):
```
AstridKery, line 212: This looks great! One thing I forgot before though, if someone adds data to the database, right now this is going to run on all of the data, including the parts which were already created.

I think you should be able to filter the Data objects by those that don't have any DataLabelSimilarityPairs objects, but let me know if that causes any trouble

backend/django/core/views/api_annotate.py
```diff
@@ -470,9 +475,11 @@ def data_admin_table(request, project_pk):
             reason = "Skipped"
 
         serialized_data = DataSerializer(d.data, many=False).data
+        print(serialized_data)
```
AstridKery, line 478: Leftover print statement

frontend/src/components/DataCard/index.jsx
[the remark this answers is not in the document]
AstridKery: Screen Shot 2021-09-02 at 11 18 23 AM

After further testing, realized this happens when the descriptions are null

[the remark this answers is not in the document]
dsteedRTI: addressed

backend/django/core/utils/util.py
AstridKery: This comment is probably resolved, but a moot point since we're moving the suggestions to another MR

backend/django/core/utils/util.py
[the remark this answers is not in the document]
AstridKery: Closing this discussion to be resumed on another MR

backend/django/core/views/api_admin.py
```diff
@@ -323,3 +325,44 @@ def heat_map_data(request, project_pk):
...
+
+    project = Project.objects.get(pk=project_pk)
+
+    total_data_objs = project.data_set.all()
+
+    final_data_objs = list(
+        DataLabel.objects.filter(data__project=project, data__irr_ind=False)
+    )
+
+    stuff_in_queue = DataQueue.objects.filter(queue__project=project)
```
AstridKery, line 350: One final thing for this (promise). The way queues work in SMART is pretty complicated. Since SMART is based on this idea that we want to label some data before others, queues are created for assignments to be pulled from.

There are three types of queues:
- admin: this is the one data are put in when they need an admin to look at them
- IRR: this is where data which is IRR and should be assigned first goes
- normal: this is where data which is not IRR and should be assigned first goes

So basically, data being in a queue doesn't mean it's been assigned yet, so we don't want to exclude all queues here. What we do want to exclude is everything in the admin queue (since that's at least been skipped or something), and all assigned data. Also anything in the IRRLog since that indicates someone has labeled/skipped that data.

All this to say I think the following change is needed.

```
    stuff_in_irrlog = IRRLog.objects.filter(data__project=project).values_list(
        "data__pk", flat=True
    )
    stuff_in_queue = DataQueue.objects.filter(queue__project=project, queue__type="admin").values_list(
        "data__pk", flat=True
    )
    recycle_ids = RecycleBin.objects.filter(data__project=project).values_list(
        "data__pk", flat=True
    )
    assigned_ids = AssignedData.objects.filter(data__project=project).values_list(
        "data__pk", flat=True
    )
    unlabeled_data_objs = list(
        project.data_set.filter(datalabel__isnull=True)
        .exclude(id__in=stuff_in_queue)
        .exclude(id__in=stuff_in_irrlog)
        .exclude(id__in=recycle_ids)
        .exclude(id__in=assigned_ids)
    )
```

AstridKery:
thanks @dsteedRTI ! Just address this last discussion and I think we're good.

backend/django/core/views/api_admin.py
dsteedRTI: thanks for this! just added it

AstridKery: Ok! I think this looks good

AstridKery (approved):

PR #114, opened 2021-09-14
Dedup on metadata

AstridKery:
This MR makes the following changes:

* when uploading data, the user has the option to select their method for deduplication. The options for this are:
  * Text: dedup based on text only (default, and how it worked before)
  * Text and all metadata (all additional fields are used)
  * Text and select metadata (user must then provide a string)
* on the details page, the user's decisions for this choice, as well as the list of all metadata fields has been added
* When downloading labeled data, the associated metadata fields are now included (since without them it would be impossible to tell data with the same text but different metadata apart.


dsteedRTI (approved):
Tested this a couple times, after pulling the recent changes, everything worked as expected!

PR #116, opened 2021-09-16
change metadata to use table insert

AstridKery:
Trying to speed up the metadata creation task. I'm leaving in the print statements for debugging.

Basically, I noticed when uploading a large file that the data object creation finishes pretty quickly (under a minute) but then the metadata object creation lags for a really long time. I updated the metadata methodology to get around some major causes of slowness, but I'm running into some weird issues. 

Basically, it seems like it's lagging right after the print `print("There are", len(df["hash"].unique()), "unique hashes")` which would suggest that the issue is the dataframe creation. However, I've also seen that run pretty fast and then the lag is after that. It's not consistent. 

I have a few theories as to what's happening. 

1. Those other times were flukes and it really is just some slow operations
2. Something is interrupting the function and causing the whole thing to stall (I'm looking at Redis maybe? we certainly get warnings over a slowdown and the next log after that last print is always from Redis doing auto saving)
3. There's a weird delay where the data objects which were just added to the table aren't actually available yet causing a delay

@peridoteagle , would you mind taking a look?

### DIFF
--- a/backend/django/core/utils/util.py
+++ b/backend/django/core/utils/util.py
@@ -195,31 +195,19 @@ def create_labels_from_csv(df, project):
 
 def create_metadata_objects_from_csv(df, project):
     """Insert metadata objects into database."""
-    print("IN THE FUNCTION")
     metadataFields = MetaDataField.objects.filter(project=project)
-    print("Getting the IDs from the hashes")
-    print("There are", len(df["hash"].unique()), "unique hashes")
     data_objects = pd.DataFrame(
-        list(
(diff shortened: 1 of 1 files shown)


AstridKery:
@peridoteagle I think I've fixed the problem! It turns out the call to filter was slowing things down dramatically. It's way faster to just pull all the data and drop the extra stuff in the pandas merge.

peridoteagle:
Nice work! This looks good - good find on the issue with filter slowing things down

PR #117, opened 2021-09-18
String grouper computations

AstridKery:
Hi @dsteedRTI ! I messed with this branch a bit and was able to speed it up. When you get the chance, would you mind making the changes you had before so the label suggestions are clickable? Thanks!


dsteedRTI (approved):

AstridKery:
setting back to WIP until we figure out how to handle how slow it is with the 300k dataset

PR #118, opened 2021-09-21
update project status on details page

dsteedRTI:
This updates the details page to have more accurate project status information

### DIFF
--- a/backend/django/core/templates/projects/detail.html
+++ b/backend/django/core/templates/projects/detail.html
@@ -48,7 +48,9 @@
             <div class="panel-body">
               <p>Fully Labeled: <span id="fully-labeled"></span></p>
               <p>Unlabeled and Unassigned: <span id="unlabeled-unassigned"></span></p>
-              <p>Other (Partially Labeled/IRR/Skipped): <span id="other-labeled"></span></p>
+              <p>Awaiting Adjudication: <span id="awaiting-adjudication"></span></p>
+              <p>Recycle Bin: <span id="recycled"></span></p>
+              <p>Partially Labeled/IRR): <span id="other-labeled"></span></p>
             </div>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


backend/django/core/templates/projects/detail.html
```diff
@@ -48,7 +48,9 @@ <h5 class="panel-title">
             <div class="panel-body">
               <p>Fully Labeled: <span id="fully-labeled"></span></p>
               <p>Unlabeled and Unassigned: <span id="unlabeled-unassigned"></span></p>
-              <p>Other (Partially Labeled/IRR/Skipped): <span id="other-labeled"></span></p>
+              <p>Awaiting Adjudication: <span id="awaiting-adjudication"></span></p>
+              <p>Recycle Bin: <span id="recycled"></span></p>
+              <p>Partially Labeled/IRR): <span id="other-labeled"></span></p>
```
AstridKery, line 53: Two notes about the other-labeled group:

1. There's a ) without a partner
2. This group also includes assigned data which has been sent to someone but not labeled yet

backend/django/core/views/api_admin.py
```diff
@@ -362,15 +362,15 @@ def get_project_status(request, project_pk):
     )
     unlabeled_data_objs = list(
         project.data_set.filter(datalabel__isnull=True)
-        .exclude(id__in=stuff_in_queue)
         .exclude(id__in=stuff_in_irrlog)
-        .exclude(id__in=recycle_ids)
```
AstridKery, line 367: Is there a reason we are including recycled items in the unlabeled count? I think that might get confusing since we explicitly have a recycled count. In addition, if we include recycled items here, then they are subtracted twice in the "other-labeled" calculation

backend/django/core/views/api_admin.py
```diff
@@ -362,15 +362,15 @@ def get_project_status(request, project_pk):
     )
     unlabeled_data_objs = list(
         project.data_set.filter(datalabel__isnull=True)
-        .exclude(id__in=stuff_in_queue)
```
AstridKery, line 365: Same question as recycled. By removing this exclude call, we add the admin items to this total.

dsteedRTI:
fixed

PR #120, opened 2021-09-22
EWDCODE-59 expand select width

jdpdev:
Also adjusts the styling of the card to prevent long text from pushing the suggestion box down

### DIFF
--- a/frontend/src/components/DataCard/index.jsx
+++ b/frontend/src/components/DataCard/index.jsx
@@ -117,7 +117,7 @@ class DataCard extends React.Component {
                                     }}
                                     variant="info"
                                 >
-                                    Ajudicate
+                                    Adjudicate
                                 </Button>
                             </OverlayTrigger>
                         </ButtonToolbar>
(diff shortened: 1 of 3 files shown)


dsteedRTI:
This looks great! Nailed it! One small think I'd like your thoughts on to:

I'm thinking we might need a media query or something to handle smaller screens, I think the minWidth of 400px is causing it to look like this on smaller browsers?
Screen Shot 2021-09-22 at 12 33 16 PM

dsteedRTI:
Also, this is from the other MR I think that was already merged?
[https://github.com/RTIInternational/SMART/pull/119](https://github.com/RTIInternational/SMART/pull/119)

The "Skip" button is actually on several pages, such as the History card. So we need to make sure to rename all instances of that button with "Adjudicate" in order to avoid confusion

jdpdev:
@dsteedRTI Adjusted select width based on screen width. Ajudicate button won't wrap when there is a select, but when there are buttons a margin is added to the top for spacing when wrapping

dsteedRTI (approved):
Awesome, looks great. You killed it!

PR #122, opened 2021-09-22
add label dropdown description to placeholder on history data cards

dsteedRTI:
This adds the current label and description to the placeholder on the dropdown/search for data cards on the history tab


AstridKery (approved):
Screen Shot 2021-09-23 at 4 36 06 PM

Not really a big issue, but realizing that sometimes the text can go out of bounds. Is there any way to truncate the text in the dropdown library?

dsteedRTI:
Good catch! Just added that

AstridKery:
Looks good!

PR #124, opened 2021-09-23
EWDCODE-46 chart total user labels

jdpdev:
Screen Shot 2021-09-23 at 2 44 40 PM

### DIFF
--- a/backend/django/core/views/api_admin.py
+++ b/backend/django/core/views/api_admin.py
@@ -56,7 +56,7 @@ def label_distribution(request, project_pk):
         if label_count > 0:
             user_labels.append({"x": u.__str__(), "y": label_count})
 
-    dataset = [{"key": "labels", "values": user_labels}]
+    dataset = [{"key": "Count", "values": user_labels}]
 
     return Response(dataset)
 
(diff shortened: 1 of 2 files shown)


dsteedRTI (approved):
This one looks great too, you're killing it! It loads faster for me than the previous one, which was important. I'm approving but going to also wait and let @AstridKery take a look at this one

backend/django/core/views/api_admin.py
```diff
@@ -48,26 +48,15 @@ def label_distribution(request, project_pk):
...
+    
     for u in users:
-        other_count = 0
-        for label in labels[5:]:
-            other_count += DataLabel.objects.filter(profile=u, label=label).count()
-        if other_count > 0:
-            other_values.append({"x": u.__str__(), "y": other_count})
-    if other_values != []:
-        dataset.append({"key": "other labels", "values": other_values})
+        label_count = DataLabel.objects.filter(profile=u, label__in=labels).count()
```
AstridKery, line 54: I like the updated chart! I do have a note about the backend function. Since we're not splitting the results by label anymore, we can actually just use the query `label_count = DataLabel.objects.filter(profile=u, label__project=project_pk).count()`. This way we can remove all mentions of `labels` entirely which should be a bit faster.

jdpdev: fixed in e276187

PR #125, opened 2021-09-27
add umbrellas/folders with string

dsteedRTI:


AstridKery:
So external from these changes, we need the following:

1. Merge in npsas_main to get this branch up to date
2. Change the permissions level for `get_project_status` from admin to coder by changing the permission class tag (see api_annotate for examples).

backend/django/core/templates/projects/list.html
```diff
@@ -7,32 +7,88 @@
...
-                {% else %}
-                <td><p>No Labeled Data to Download</p></td>
-                {% endif %}
-              {% endif %}
-            {% endif %}
-          </tr>
-          {% endfor %}
-        </tbody>
-      </table>
+      {% for potential_umbrella_project in object_list %}
```
AstridKery, line 10: So... kinda a dumb question. This is obviously working, but I'm trying to figure out *why* it's working. 

Basically, if there are three projects in folder "testing" then I'm wondering why we don't end up with three "testing" panels with the three projects inside? The outer loop goes over all the projects, so in theory the inner loop should run all three times as well....

dsteedRTI: Yea, you're totally right. Basically, at the bottom I wrote some js in a script to "cleanup" and get rid of any extra panels

backend/django/core/templates/projects/update/umbrella.html
```diff
@@ -0,0 +1,35 @@
...
+           <h1>Add Project To Folder</h1>
+           <h3>Description</h3>
+           <p>To help organize your labeling projects, you can assign projects to folders to organize your project list.</p>
+           <ul class="list-group">
+             <li class="list-group-item"><strong>Admins</strong> are able to update the project's folder to one they have created.</li>
+           </ul>
+           <h3>Instructions</h3>
+           <p>Select a folder from the dropdown field below to add your project. If you want to add the project to a folder that is not on the list, contact the folder's owner.</p>
+           {{ form.errors.as_ul }}
+           {{ permissions.management_form }}
```
AstridKery, line 23: There are a few places here with `permissions.` which needs to be updated

backend/django/core/templates/projects/list.html
AstridKery: Oh nevermind. I see you've added a function at the bottom here which removes duplicates

backend/django/core/templates/projects/update/umbrella.html
```diff
@@ -0,0 +1,35 @@
...
+         <form action="." method="post" enctype="multipart/form-data">
+           {% csrf_token %}
+           <h1>Add Project To Folder</h1>
+           <h3>Description</h3>
+           <p>To help organize your labeling projects, you can assign projects to folders to organize your project list.</p>
+           <ul class="list-group">
+             <li class="list-group-item"><strong>Admins</strong> are able to update the project's folder to one they have created.</li>
+           </ul>
+           <h3>Instructions</h3>
+           <p>Select a folder from the dropdown field below to add your project. If you want to add the project to a folder that is not on the list, contact the folder's owner.</p>
```
AstridKery, line 21: I'd also add "or type a new folder name in the text box to create a new folder and add this project to it." 

That way users know that they can create new folders in this update page.

AstridKery (changes requested):
Just a few changes, but otherwise this looks great. Thanks Durk!

AstridKery (approved):
Looks good! I think we can merge this.

PR #129, opened 2021-11-18
Update dependency management system

AstridKery:
This branch makes the following changes to the SMART repo:

1. Updates the package.json file to use the `^` for all versions. This means that bug fixes to versions will be incorporated.
2. Overhauls the backend versioning to use pip-tools (https://github.com/jazzband/pip-tools). This is a dependency manager that generates a requirements.txt file with all dependencies generated and pinned. The upshot of this is that we really just need to maintain the primary libraries in `requirements.in`.  
3. Fixes a few broken tests that were leftover from previous changes

I've assigned you both as reviewers because I'd like to make sure everyone is able to follow the backend dependency update steps in the README and it all makes sense.


dsteedRTI (approved):

dsteedRTI:
The pip-compile commands worked; however, I got the following warnings when running:

`WARNING: isort 5.10.1 does not provide the extra 'requirements'`
`WARNING: The candidate selected for download or install is a yanked version: 'black' candidate (version 21.11b0 at https://files.pythonhosted.org/packages/3d/ad/1cf514e7f9ee4c3d8df7c839d7977f7605ad76557f3fca741ec67f76dba6/black-21.11b0-py3-none-any.whl#sha256=0b1f66cbfadcd332ceeaeecf6373d9991d451868d2e2219ad0ac1213fb701117 (from https://pypi.org/simple/black/) (requires-python:>=3.6.2))
  Reason for being yanked: Broken regex dependency. Use 21.11b1 instead.`

AstridKery:
Ok I've made a couple of updates to address the warnings from some of the small version changes. @peridoteagle when you get time, I'd appreciate your review before this is merged, so we can make sure everyone is on the same page about these changes.

peridoteagle:
Just to add that I also ran the pip-compile command and did not see any warnings

PR #130, opened 2021-11-19
prototype string comparison client side on the fly

dsteedRTI:
This prototypes trying the string comparisons for labels to card text on the client side (fixing our problem with database storage taking forever with large amounts of data and labels.)


frontend/src/components/DataCard/SuggestedLabels.jsx
```diff
@@ -1,15 +1,21 @@
...
 
 export default function SuggestedLabels({ card, labels }) {
     if (labels.length <= 5) {
         return null;
     }
 
+    const labelStrings = labels.map(label => `${label.name} (${label.description})`);
+
+    let matches = stringSimilarity.findBestMatch(card.text.text, labelStrings);
+    matches = matches.ratings.sort((rating1, rating2) => parseFloat(rating2.rating) - parseFloat(rating1.rating)).slice(0, 5);
```
AstridKery, line 12: Would you mind adding a comment here to explain how this works?

frontend/src/components/DataCard/SuggestedLabels.jsx
```diff
@@ -1,15 +1,21 @@
...
+
     return (
         <div className="suggestions">
             <h4>Suggested Labels</h4>
-            {card.text.similarityPair.slice(0, 5).map((opt, index) => (
```
AstridKery, line 11: I think this method works really well. If it doesn't rely on any of the string similarity stuff on the backend, then we should remove all of that stuff in this MR.

frontend/src/components/DataCard/SuggestedLabels.jsx
```diff
@@ -1,15 +1,21 @@
...
-                <div key={index + 1} className="">{index + 1}. {opt.split(':')[0]}</div>
+            {matches.map((match, index) => (
+                <div key={index + 1} className="">{index + 1}. {match.target}</div>
```
AstridKery, line 18: I seem to recall that we wanted these suggestions to be buttons. Would it be straightforward to have them be clickable as labels?

frontend/src/components/DataCard/SuggestedLabels.jsx
dsteedRTI: Sure!

`labelStrings` is going to be an array of strings with the format: "{label name} {label description}"

The `stringSimilarity.findBestMatch` takes a string and an array of strings to preform the similarity comparisons and is saved in the variable `matches`.

`matches` is then sorted from best rating to lowest rating (ratings are floats, so they're parsed first then compared against each other.) 

Then I'm simply splicing to only get the first 5 results to loop through in the JSX.

frontend/src/components/DataCard/SuggestedLabels.jsx
dsteedRTI: Yea! Ok, now that we like this way better, I'll get to removing all the different backend stuff we had previously.

frontend/src/components/DataCard/SuggestedLabels.jsx
dsteedRTI: Sure, I think it should, I can also look into this.

frontend/src/components/DataCard/SuggestedLabels.jsx
AstridKery: Lol I meant please add that explanation to the code in the form of a comment so we can know what it's doing later :)

AstridKery:
Hi @dsteedRTI ! The changes you've made make sense to me. I'm seeing an error now with the history table for many-label projects, but I'm guessing this is because you haven't updated the other label methods?
Screen Shot 2021-12-02 at 2 54 26 PM

AstridKery:
Hi @dsteedRTI ! Would you mind updating this branch to add the string similarity to the other tables and fix the dependency merge conflict? Thanks!

PR #131, opened 2021-12-14
remove default batch selection from create page

dsteedRTI:
This removes the default batch option in the project create form and instead just has an input for the user to determine the batch size (as opposed to defaulting to 10 times the number of labels.) 

The default value in the input still starts at 30.


AstridKery (approved):
This works for me! I tested on 1k+ labels and it worked fine.

PR #143, opened 2022-02-08
add project status badge

dsteedRTI:
This adds project status badges to the project list page.

It moves the logic to get the project status to utils, and also changes the old details page to get the status from this new utils function

### DIFF
--- a/backend/django/core/utils/util.py
+++ b/backend/django/core/utils/util.py
@@ -14,9 +14,11 @@ from string_grouper import compute_pairwise_similarities
 
 from core import tasks
 from core.models import (
+    AssignedData,
     Data,
     DataLabel,
     DataLabelSimilarityPairs,
+    DataQueue,
     IRRLog,
     Label,
     MetaData,
@@ -24,6 +26,7 @@ from core.models import (
     Profile,
     Project,
     ProjectPermissions,
+    RecycleBin,
     TrainingSet,
 )
 from core.utils.utils_queue import fill_queue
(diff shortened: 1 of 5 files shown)


backend/django/core/utils/util.py
```diff
@@ -571,3 +574,48 @@ def get_labeled_data(project):
...
+    )
+
+    return {
+        "adjudication": len(stuff_in_queue),
+        "assigned": len(assigned_ids),
+        "final": len(final_data_objs),
+        "recycled": len(recycle_ids),
+        "total": len(total_data_objs),
+        "unlabeled": len(unlabeled_data_objs),
+        "badge": f"{len(total_data_objs) - len(unlabeled_data_objs)}/{len(unlabeled_data_objs)}",
```
AstridKery, line 620: So a few notes:

* we should probably not count things in the recycle bin as data to be labeled. It's basically been "thrown out" from the project 
* the numerator as-is includes data which is assigned or partially labeled. So someone might see 100/100 and think everything is labeled when really it's not. I would suggest we change this to just count the number of data with a final label.

AstridKery: What do you think of the second bullet? I'm just anticipating someone might get confused.

backend/django/core/utils/util.py
```diff
@@ -575,3 +577,48 @@ def get_labeled_data(project):
...
+    total_data_objs = project.data_set.all()
+
+    final_data_objs = list(
+        DataLabel.objects.filter(data__project=project, data__irr_ind=False)
+    )
+
+    stuff_in_irrlog = IRRLog.objects.filter(data__project=project).values_list(
+        "data__pk", flat=True
+    )
+    stuff_in_queue = DataQueue.objects.filter(
```
AstridKery, line 599: `NameError: name 'DataQueue' is not defined`

PR #148, opened 2022-02-18
Npsas main checkboxes v2

peridoteagle:
This commit:
* Makes the default for Active Learning unchecked
* Default for IRR is unchecked
* Fixes a bug with data upload for CSVs and XLSXs


backend/django/core/templates/projects/create/create_wizard_advanced.html
```diff
@@ -58,7 +58,7 @@ <h5 class="panel-title">
                   <div id="choose_classifier_box">
                     <div id="use_model_div">
                       <p>Train and display results for a model: {{ wizard.form.use_model }}</p>
-                      <p>NOTE: If you uncheck this option, you cannot use Active Learning</p>
```
AstridKery, line 61: Kind of a nitpick, but I'm noticing when I check the model option, the active learning option pops up above instead of below, which made sense originally since you might disable active learning first. In this new "opt in" workflow, we probably want the model section to go above the active learning section.

peridoteagle: Agreed - great idea. Reordered in commit

AstridKery (approved):
Looks good to me!

PR #151, opened 2022-02-27
update data forms so that it loads with the correct initial values

AstridKery:
[EWDCODE-106](https://jira.rti.org/browse/EWDCODE-106) Adds some fixes to the data page so the forms initialize with the correct items hidden/shown after incorrect input or going back a page.


dsteedRTI (approved):
Ok this is awesome! This will definitely improve the user experience especially for onboarding newer users.

PR #152, opened 2022-03-02
handle more than 2000 labels when updating descriptions

dsteedRTI:
The default max size of a form is 2,000 according to the Django docs, this adds a max limit of 10,000 so that we can update projects with many labels.

Django docs: https://docs.djangoproject.com/en/4.0/topics/forms/formsets/#limiting-the-maximum-number-of-instantiated-forms

### DIFF
--- a/backend/django/core/forms.py
+++ b/backend/django/core/forms.py
@@ -293,7 +293,12 @@ LabelFormSet = forms.inlineformset_factory(
     absolute_max=10000,
 )
 LabelDescriptionFormSet = forms.inlineformset_factory(
-    Project, Label, form=LabelDescriptionForm, can_delete=False, extra=0
+    Project,
+    Label,
+    form=LabelDescriptionForm,
+    can_delete=False,
+    extra=0,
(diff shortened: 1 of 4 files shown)


backend/django/core/forms.py
```diff
@@ -293,7 +293,12 @@ def __init__(self, *args, **kwargs):
     absolute_max=10000,
 )
 LabelDescriptionFormSet = forms.inlineformset_factory(
-    Project, Label, form=LabelDescriptionForm, can_delete=False, extra=0
+    Project,
+    Label,
+    form=LabelDescriptionForm,
+    can_delete=False,
+    extra=0,
+    absolute_max=10000,
```
AstridKery, line 301: Do we say anywhere that 10k is the max for labels? If it's not on the Label creation form we should add it.

dsteedRTI: Good thinking, added.

AstridKery (approved):
Looks good!

PR #154, opened 2022-03-07
enable skew to handle large projects

dsteedRTI:
This limits the number of unlabeled data to 50 on the skew table to prevent large projects from lagging out. It also adds a search input on top of the table to search through all of the unlabeled data for the project. A new API endpoint is used to handle searches.

### DIFF
--- a/backend/django/core/views/api_annotate.py
+++ b/backend/django/core/views/api_annotate.py
@@ -428,19 +428,16 @@ def data_unlabeled_table(request, project_pk):
     Returns:
         data: a list of data information
     """
-    unlabeled_data = get_unlabeled_data(project_pk)
-    data = []
-    for d in unlabeled_data:
-        if len(data) < 50:
-            serialized_data = DataSerializer(d, many=False).data
-
-            temp = {
-                "Text": serialized_data["text"],
-                "metadata": serialized_data["metadata"],
-                "ID": d.id,
-            }
-            data.append(temp)
-
+    unlabeled_data = get_unlabeled_data(project_pk)[:50]
+    serialized_data = DataSerializer(unlabeled_data, many=True).data
+    data = [
+        {
+            "Text": d["text"],
+            "metadata": d["metadata"],
+            "ID": d["pk"],
+        }
+        for d in serialized_data
+    ]
     return Response({"data": data})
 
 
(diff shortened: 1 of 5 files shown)


backend/django/core/views/api_annotate.py
```diff
@@ -427,29 +428,50 @@ def data_unlabeled_table(request, project_pk):
...
-        project.data_set.filter(datalabel__isnull=True)
-        .exclude(id__in=queued_ids)
-        .exclude(id__in=recycle_ids)
-    )
+    return Response({"data": data})
+
+
+@api_view(["GET"])
+@permission_classes((IsAdminOrCreator,))
+def search_data_unlabeled_table(request, project_pk):
```
AstridKery, line 449: Thanks for setting all of this up! I was looking at this and wanted to see if I could speed it up by using some of the built-in django operations and it seems to work. 

Basically, if you use the built-in "icontains" filter which is a case-insensitive text match and dump everything in the serializer together you can get the same output with the code below. I *think* this is faster? 

```
    unlabeled_data = get_unlabeled_data(project_pk)
    text = request.GET.get("text")
    unlabeled_data = unlabeled_data.filter(text__icontains=text.lower())
    serialized_data = DataSerializer(unlabeled_data, many=True).data
    data = [{
        "data": d["text"],
        "metadata": d["metadata"],
        "id": d["pk"],
        "project": project_pk,
    } for d in serialized_data]
    return Response({"data": data})
```

AstridKery, line 449: Thanks for setting all of this up! I was looking at this and wanted to see if I could speed it up by using some of the built-in django operations and it seems to work. 

Basically, if you use the built-in "icontains" filter which is a case-insensitive text match and dump everything in the serializer together you can get the same output with the code below. I *think* this is faster? 

```
    unlabeled_data = get_unlabeled_data(project_pk)
    text = request.GET.get("text")
    unlabeled_data = unlabeled_data.filter(text__icontains=text.lower())
    serialized_data = DataSerializer(unlabeled_data, many=True).data
    data = [{
        "data": d["text"],
        "metadata": d["metadata"],
        "id": d["pk"],
        "project": project_pk,
    } for d in serialized_data]
    return Response({"data": data})
```

backend/django/core/views/api_annotate.py
```diff
@@ -427,29 +428,50 @@ def data_unlabeled_table(request, project_pk):
     Returns:
         data: a list of data information
     """
-    project = Project.objects.get(pk=project_pk)
+    unlabeled_data = get_unlabeled_data(project_pk)
+    data = []
```
AstridKery, line 432: Same as with the `search_data_unlabeled_table` but simpler. I think we can get this effect faster by moving the serialization to bulk with something like this.

```
    unlabeled_data = get_unlabeled_data(project_pk)[:50]
    serialized_data = DataSerializer(unlabeled_data, many=True).data
    data = [{
        "Text": d["text"],
        "metadata": d["metadata"],
        "ID": d["pk"],
    } for d in serialized_data]
    return Response({"data": data})
```

AstridKery, line 432: Same as with the `search_data_unlabeled_table` but simpler. I think we can get this effect faster by moving the serialization to bulk with something like this.

```
    unlabeled_data = get_unlabeled_data(project_pk)[:50]
    serialized_data = DataSerializer(unlabeled_data, many=True).data
    data = [{
        "Text": d["text"],
        "metadata": d["metadata"],
        "ID": d["pk"],
    } for d in serialized_data]
    return Response({"data": data})
```

dsteedRTI: feels faster to me, thanks!

AstridKery (approved):
Looks good!

Issue #155, opened 2022-03-08
WebpackError at /accounts/login/


gundarskorats:
OS: Ubuntu 20.04 
Problem: Followed installation instructions successfully (no errors during building process), but encountered with WebpackError at /accounts/login/.
Any suggestions how to solve this?

<h1 style="padding: 0px; margin: 0px; font-weight: normal; color: rgb(0, 0, 0); font-family: sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">WebpackError at /accounts/login/</h1><pre class="exception_value" style="padding: 0px; margin: 10px 0px; font-size: 1.5em; white-space: pre-wrap; font-family: sans-serif; color: rgb(87, 87, 87); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">            ModuleBuildError in 
            Module build failed (from ../node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ../node_modules/sass-loader/dist/cjs.js):
SassError: Invalid CSS after "...     floor(math": expected expression (e.g. 1px, bold), was ".div($grid-gutter-w"
        on line 369 of node_modules/bootstrap-sass/assets/stylesheets/bootstrap/_variables.scss
        from line 8 of node_modules/bootstrap-sass/assets/stylesheets/_bootstrap.scss
        from line 2 of src/styles/_base.scss
        from line 5 of src/styles/_includes.scss
        from line 2 of src/styles/smart.scss
&gt;&gt; vbar-padding-horizontal:        floor(math.div($grid-gutter-width, 2)) !defa
   ------------------------------------------^

    at processResult (/code/node_modules/webpack/lib/NormalModule.js:753:19)
    at /code/node_modules/webpack/lib/NormalModule.js:855:5
    at /code/node_modules/loader-runner/lib/LoaderRunner.js:399:11
[...]

PR #156, opened 2022-03-08
adding updated yarn.lock to fix webpack error

AstridKery:
Created to close issue #155 . When the frontend dependency file package.json was updated in master, the yarn.lock file was not updated at the same time, leading to a webpack error in building. 

To test this fix, please rebuild your frontend docker container with the tag `--no-cache` to ensure the new version is actually used.


dsteedRTI (approved):
Works well for me

PR #162, opened 2022-03-30
replicating dev in prod

hardingalexh:
This merge request adds a functional production docker-compose setup, with the frontend emulating the dev setup. Running the `build-prod.sh` bash script from `/envs/prod` should bring up a functional tool at port 80.

Lots more to say about the exploration required to come to this solution - let's chat about that soon.


AstridKery:
Thanks @hardingalexh! This is running for me now. The only thing I'd recommend is making build-prod.sh executable. So it can be called with ./

PR #163, opened 2022-03-30
Use Local, Pre-trained model

dsteedRTI:
This adds a local modal for our embeddings calculations using the `.csv` file @peridoteagle created with some specific abbreviations and matching text pairs.

If we want to make new models in the future, I made this [Github repo](https://github.com/dsteedRTI/csv-to-embeddings-model) that will take a `.csv` file, format it, and create a new pre-trained model.


backend/django/core/views/smart_embeddings_model/README.md
```diff
@@ -0,0 +1,87 @@
...
+- sentence-transformers
+- feature-extraction
+- sentence-similarity
+---
+
+# {MODEL_NAME}
+
+This is a [sentence-transformers](https://www.SBERT.net) model: It maps sentences & paragraphs to a 384 dimensional dense vector space and can be used for tasks like clustering or semantic search.
+
+<!--- Describe your model here -->
```
AstridKery, line 13: Seeing these comments here. It looks like this is a template. Do we want to fill these in?

dsteedRTI: Yea we can fill the README out if we want!

PR #164, opened 2022-04-06
WIP: Order data by newest or oldest

AstridKery:
This branch adds the "newest" or "oldest" options for ordering the data passed out in SMART. This required the following updates:

* saving the time data was uploaded as a property of data. Add it to the cards being shown on the frontend.
* modifying the project ordering logic which originally assumed "random" was the only non-active-learning ordering possible. 
* significantly updating the advanced settings page to make these new updates make sense
* update the logic in fill_queue and check_and_trigger_model to handle these other ordering options

This update touched a lot of places, and concern about newly introduced bugs is high. Please take a look, and let's discuss some new backend tests to verify the functionality. Thanks!


AstridKery:
Added a fix so that it doesn't assume that someone is using a model for their ordering if they have one. Also added a chec_and_trigger_model call to the card deck request since if something went wrong with their annotating (like clicking the button twice) then it might not have been called. That function is used to refill the queue, so the user could be stuck without any cards forever.

PR #169, opened 2022-05-06
hide progress bar, move adjudicate button

dsteedRTI:
Hides the progress bar (I just commented it out and replace with a React Fragment so we have the code there in case we need it in the future.)

Moves the adjudicate button to the top right of the card.

### DIFF
--- a/frontend/src/components/DataCard/CardData.jsx
+++ b/frontend/src/components/DataCard/CardData.jsx
@@ -1,15 +1,45 @@
 import React from "react";
+import {
+    Button,
+    Tooltip,
+    OverlayTrigger
+} from "react-bootstrap";
 
-export default function CardData({ card }) {
+export default function CardData({ card, onSkip }) {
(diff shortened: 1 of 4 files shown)


frontend/src/components/DataCard/CardData.jsx
```diff
@@ -1,9 +1,17 @@
...
+    OverlayTrigger
+} from "react-bootstrap";
 
-export default function CardData({ card }) {
+export default function CardData({ card, onSkip }) {
     return (
         <div className="cardface-info">
-            <h2>Card {card.id + 1}</h2>
+            <div className="card-title">
+                <h2>Card {card.id + 1}</h2>
```
AstridKery, line 12: So I believe this may not have been discussed before this MR was made, but since we don't have a progress bar anymore, the card numbers are probably more confusing than anything else. So we probably want to drop those too.

PR #171, opened 2022-05-13
Adjudicte descriptions

dsteedRTI:
This branch adds:
AdjudicateDescriptions - model to add reasons for why a card was skipped
Display descriptions above cards on the admin table

Summary:

When a user "adjudicates" a card, a reason/message is collected and stored in the db. When an admin views the admin table, this message is displayed on top of the card for why the card was skipped. When an admin labels the card, the message's status is set to resolved in the db (but kept, in case they want to export/keep all reasons.)


dsteedRTI:
Leaving this here, will fix it next week: Deleting projects errors out because of a foreign key constraint on the adjudication descriptions

backend/django/core/migrations/0068_adjudicate_descriptions_20220510_2033.py
```diff
@@ -0,0 +1,50 @@
...
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("message", models.TextField()),
+                ("isResolved", models.BooleanField(default=False)),
+                (
+                    "data",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE, to="core.data"
```
dsteedRTI, line 39: `CASCADE` here makes it safe to delete projects, ignore my earlier comment. I was on another branch working on a separate issue when I encountered the issue, it works fine here! My bad!

dsteedRTI:
> @dsteedRTI just checking - what is the status of this PR?

@peridoteagle this is ready just waiting on a review.

frontend/src/components/AnnotateCard/index.jsx
```diff
@@ -97,6 +97,11 @@ function drawSkipButton(card, onSkip) {
     );
 }
 
+function promptAdjudicate(card, onSkip) {
+    let message = prompt("Please enter the reasons for skipping this card:");
```
AstridKery, line 101: So I'm wondering if we could use a model for this instead?

https://react-bootstrap.github.io/components/modal/

It might make it easier to keep with the styling of the rest of the app. Otherwise my thoughts are these:

1. The reason should probably not be submittable blank. (would check on Wed with Ethan though)
2. My prompt says "localhost:800 says" which would confuse a normal user. Can we change this?
3. We'd want to change "ok" to "submit" since they are submitting a form

frontend/src/components/AdminTable/index.jsx
```diff
@@ -77,7 +86,7 @@ class AdminTable extends React.Component {
                 </p>
                 <CodebookLabelMenuContainer />
                 <ReactTable
-                    data={admin_data}
```
AstridKery, line 80: Getting something kind of weird with the admin table pagination? I can't seem to change the number of items to show anymore. Do you have this issue?
Screen Shot 2022-05-24 at 5 06 43 PM

frontend/src/components/AdminTable/index.jsx
```diff
@@ -48,16 +48,25 @@ class AdminTable extends React.Component {
...
-                        />
-                    </div>
-                )
+                Cell: row => {
+                    const potentialReasonMessage = admin_data.messages.find((message) => message["data_id"] === row.row.id);
+                    return (
+                        <div className="sub-row">
+                            {potentialReasonMessage && (
+                                <div className="adjudicate-message">
+                                    <h4>Reason:</h4>
```
AstridKery, line 57: I'd change this to "Reason for Skipping" for clarity

backend/django/core/views/api_annotate.py
```diff
@@ -685,6 +694,8 @@ def label_admin_label(request, data_pk):
         if datum.irr_ind:
             Data.objects.filter(pk=datum.pk).update(irr_ind=False)
 
+        AdjudicateDescription.objects.filter(data_id=data_pk).update(isResolved=True)
```
AstridKery, line 697: probably need to add this to the discard data call as well, so it's not hanging around if the data gets restored and put back in circulation

backend/django/core/views/api_annotate.py
```diff
@@ -555,7 +558,13 @@ def data_admin_table(request, project_pk):
         }
         data.append(temp)
 
-    return Response({"data": data})
+    messages = list(
```
AstridKery, line 561: So I see we're sending the messages separately here from the data items they go with. Then on the frontend we have to search for the correct data message each time. This isn't necessarily wrong, but I'm wondering if we'll have runtime issues with more data. Is there a reason you decided to pull them this way instead of pulling the message object in the data loop above?

frontend/src/components/AdminTable/index.jsx
dsteedRTI: Fixed

dsteedRTI:
@AstridKery updated quite a bit to this:

- Added the messages to the rest of the card's data instead of bringing them separately
- Fixed the rows issue on the admin table
- Added a modal that collects a required message before adjudicating a card
- Added a check to resolve any outstanding adjudications on a card on the discard call

backend/django/core/views/api_annotate.py
```diff
@@ -546,6 +552,12 @@ def data_admin_table(request, project_pk):
     project = Project.objects.get(pk=project_pk)
     queue = Queue.objects.get(project=project, type="admin")
 
+    messages = list(
```
AstridKery, line 555: I'm wondering if we can just move this into the data serializer like the metadata? If there's some reason it's not straightforward that's fine, but I'm worried this might get slow later.

backend/django/core/views/api_annotate.py
```diff
@@ -561,6 +573,7 @@ def data_admin_table(request, project_pk):
             "metadata": serialized_data["metadata"],
             "ID": d.data.id,
             "Reason": reason,
+            "message": [x for x in messages if x["data_id"] == d.data.id][0]["message"],
```
AstridKery, line 576: I'm getting an error here

```
backend           | [13/Jun/2022 19:12:41] "GET /api/data_admin_counts/2/ HTTP/1.1" 200 27
backend           | Internal Server Error: /api/data_admin_table/2/
backend           | Traceback (most recent call last):
backend           |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
backend           |     response = get_response(request)
backend           |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
backend           |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
backend           |   File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
backend           |     return view_func(*args, **kwargs)
backend           |   File "/usr/local/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
backend           |     return self.dispatch(request, *args, **kwargs)
backend           |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
backend           |     response = self.handle_exception(exc)
backend           |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
backend           |     self.raise_uncaught_exception(exc)
backend           |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
backend           |     raise exc
backend           |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
backend           |     response = handler(request, *args, **kwargs)
backend           |   File "/usr/local/lib/python3.8/site-packages/rest_framework/decorators.py", line 50, in handler
backend           |     return func(*args, **kwargs)
backend           |   File "/code/core/views/api_annotate.py", line 576, in data_admin_table
backend           |     "message": [x for x in messages if x["data_id"] == d.data.id][0]["message"],
backend           | IndexError: list index out of range
```

frontend/src/components/DataCard/CardData.jsx
```diff
@@ -1,16 +1,19 @@
...
 } from "react-bootstrap";
 
-export default function CardData({ card, onSkip }) {
+export default function CardData({ card, onSkip, showAdjudicate = true }) {
     return (
         <div className="cardface-info">
-            <div className="card-title" style={{ display: "flex", justifyContent: 'flex-end' }}>
-                {drawSkipButton(card, onSkip)}
-            </div>
+            {showAdjudicate && (
```
AstridKery, line 12: The adjudicate button also shows up in the discard tab currently.

frontend/src/components/DataCard/CardData.jsx
```diff
@@ -1,16 +1,19 @@
...
+import React, { Fragment, useState } from "react";
 import {
     Button,
+    Modal,
     Tooltip,
     OverlayTrigger
 } from "react-bootstrap";
 
-export default function CardData({ card, onSkip }) {
+export default function CardData({ card, onSkip, showAdjudicate = true }) {
```
AstridKery, line 9: Looks like if you adjudicate something from the history table (so it was labeled but you want to send it to an admin instead), it gets skipped but the adjudication reason isn't recorded

AstridKery:
Thanks @dsteedRTI ! Just have a few more discussions but it's looking really good!

backend/django/core/views/api_annotate.py
dsteedRTI: The reason I was thinking of keeping this separate was because the user only needs to see this data on the Skipped page. On annotate, history, etc. there's no need to fetch that data (it also won't even exist) unlike the metadata which is present on every page's cards. Thoughts?

AstridKery: Oh I was thinking we could make a serializer just for the skipped page.

AstridKery: I think for now this is fine.

PR #173, opened 2022-05-23
improve label uploading speed

dsteedRTI:
This better handles uploading tons of labels in bulk so that they don't take hours to upload (should be nearly instant.)

I also added a second nabber to go to the next step if there are more than 100 labels, so that you would not have to scroll all the way to the bottom of the page to advance to the next step.

### DIFF
--- a/backend/django/core/templates/projects/create/create_wizard_labels.html
+++ b/backend/django/core/templates/projects/create/create_wizard_labels.html
@@ -47,6 +47,14 @@
             <input class="form-control" id="{{ wizard.form.data.id_for_label }}" maxlength="30" name="{{ wizard.form.data.html_name }}" type="file" placeholder="{{ form.data.label }}" onChange="handleUpload(event)" onclick="this.value = null;" />
           </div>
           <br />
+          <div class="fixed-wiz">
+            <div class="wizard_nav_bar">
+              <button class="btn btn-info" name="wizard_goto_step" type="submit" value="project">1. Info</button>
+              <input class="btn btn-primary" type="submit" value="Next Step"/>
+              <p class="pull-right">Step {{ wizard.steps.step1 }} of {{ wizard.steps.count }}</p>
+            </div>
+          </div>
+          <br />
           <table class="table table-striped" id="form-table">
             {{ wizard.form.management_form }}
             <div class="error-messages">{{ wizard.form.non_form_errors }}</div>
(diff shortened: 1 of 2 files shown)


peridoteagle:
Thanks @dsteedRTI! Pinging @AstridKery here - this is high priority for Donna and wanted to check if you might be able to review

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -104,25 +112,29 @@ <h3>Instructions</h3>
...
+                  <div class="error-messages"></div>
+                  <input type="text" name="label_set-${index}-name" class="form-control" id="id_label_set-${index}-name" value="${label[0]}"></input>
+                </td>
+                <td>
+                  <div class="error-messages"></div>
+                  <textarea name="label_set-${index}-description" cols="40" rows="5" class="form-control" id="id_label_set-${index}-description" value="${label[1]}">${label[1]}</textarea>
+                </td>
+                <td>
+                  <div class="error-messages"></div>
+                  <a class="delete-row" href="javascript:void(0)" style="">remove label</a></td>
```
AstridKery, line 130: So when I use this to add a bunch of labels, the remove label button is no longer working

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -104,25 +112,29 @@ <h3>Instructions</h3>
...
-                if($('.formset_row_label').eq(index - 1).length > 0) {
-                  $('.formset_row_label').eq(index - 1).css('display', '');
-                  $('.formset_row_label').eq(index - 1).find('input[type=text]').eq(0).val(label[0]);
-                  if(label[1].length > 0) {
-                    $('.formset_row_label').eq(index - 1).find('textarea').eq(0).val(label[1]);
-                  } else {
-                    $('.formset_row_label').eq(index - 1).find('textarea').eq(0).val('');
-                  }
-                } else {
-                  $(':contains("add label")').click();
```
AstridKery, line 118: So when we have a bunch of labels add by csv, the "add label" button seems to go away at the bottom. However I think this is also true of the current version in npsas_main. Do we know why this happens?

dsteedRTI: Swapping out setting the HTML with `prepend()` from JQuery did the trick. I think the add button is part of the table, so when the HTML was being overwritten that was also removed.

backend/django/core/templates/projects/create/create_wizard_labels.html
dsteedRTI: I wrote a new function to handle deleting labels when a csv is used, hopefully that fixes it!

dsteedRTI:
Ok, fingers crossed haha, but I pushed some changes that should fix adding/deleting labels after a CSV file has been used.

backend/django/core/templates/projects/create/create_wizard_labels.html
AstridKery: Cool! So I'm breaking things when I do the following:

1. Make a project and uploade a huge label file
2. delete the first label
3. hit continue

It seems to clear the label, but then gives an error that it's empty?
Screen Shot 2022-05-25 at 10 46 48 AM

Also the add label button vanishes again lol.
Screen Shot 2022-05-25 at 10 47 40 AM

AstridKery: I'll note that before I did 2 and 3, the add label button was indeed there

dsteedRTI: It was some weird Django requirement that you can't delete labels 1 and 2 but have to use a hidden input and turn them off. Only applied to those two. I added a fix to account for that so hopefully it works smoothly now.

AstridKery: So it looks like currently if you delete one of the first or second ones then both are removed.

The behavior to replicate:

1. load in the cip_codes.csv file
2. delete either the first or second label
3. Go to the next page
4. Go back to the previous page
5. wait for it to finish loading (there's an odd in-between page with a delete column too. hmmm)

PR #177, opened 2022-05-26
Npsas metadata and deuplication tests

schittarath3-rti:
Adds tests for metadata uploading and deduplication of data based on text, all metadata, and select metadata


AstridKery:
@dsteedRTI are you comfortable taking a first pass? I think the key with this one is to think of the different options the user has when creating a project in terms of metadata and deduping, and check that we test them thoroughly.

dsteedRTI:
@AstridKery these looked good to me on a first pass

peridoteagle:
Pinging @AstridKery - let me know if these look good to merge

AstridKery (approved):
These look great! Thanks so much!

PR #178, opened 2022-06-06
add css to fix codebook modal

dsteedRTI:
I noticed while attempting to make a modal for the adjudicate descriptions that our modal components aren't working. This adds css to fix them again. (The codebook menu modal on the annotate card page was not displaying.)


AstridKery:
This is working for me now. Thanks for catching this!

PR #179, opened 2022-06-07
Cached embeddings for faster label suggestions

schittarath3-rti:
Cached embeddings, that were originally requested for every card, for label suggesting


dsteedRTI (approved):
This is great! Well done.

PR #180, opened 2022-06-08
slugify folder names on project listing

dsteedRTI:
Uses Django's built-in [slugify](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#slugify) util to slugify the folder names on the project listing. This solves the issue where folders with certain characters such as `:` or `^` were not opening properly.

**Note:** if two folders have the same characters minus the certain characters such as `:` or `^` one of them will not display.

For example:

- `folder2022` and `folder-2022` works ✅
- `folder2022` and `folder:2022` does not work ❌

### DIFF
--- a/backend/django/core/templates/projects/list.html
+++ b/backend/django/core/templates/projects/list.html
@@ -9,15 +9,15 @@
     <div class="list-group">
       {% for potential_umbrella_project in object_list %}
         {% if potential_umbrella_project.umbrella_string  %}
-        <div id="umbrella-{{potential_umbrella_project.umbrella_string}}" class="panel panel-default">
+        <div id="umbrella-{{potential_umbrella_project.umbrella_string|slugify}}" class="panel panel-default">
           <div class="panel-heading">
             <h5 class="panel-title">
-              <a data-toggle="collapse" href="#{{potential_umbrella_project.umbrella_string.split|join:"-"}}" class="accordion-toggle collapsed collapsed">
+              <a data-toggle="collapse" href="#{{potential_umbrella_project.umbrella_string|slugify}}" class="accordion-toggle collapsed collapsed">
                 {{ potential_umbrella_project.umbrella_string }}
               </a>
             </h5>
           </div>
-          <div id="{{potential_umbrella_project.umbrella_string.split|join:"-"}}" class="panel-collapse collapse">
+          <div id="{{potential_umbrella_project.umbrella_string|slugify}}" class="panel-collapse collapse">
             <div class="panel-body">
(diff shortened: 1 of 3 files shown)


backend/django/core/templates/projects/list.html
```diff
@@ -9,15 +9,15 @@ <h1>Projects</h1>
     <div class="list-group">
       {% for potential_umbrella_project in object_list %}
         {% if potential_umbrella_project.umbrella_string  %}
-        <div id="umbrella-{{potential_umbrella_project.umbrella_string}}" class="panel panel-default">
+        <div id="umbrella-{{potential_umbrella_project.umbrella_string|slugify}}" class="panel panel-default">
```
AstridKery, line 12: Screen Shot 2022-06-13 at 3 21 37 PM
So if someone makes a project with only special characters, the issue still remains. Thoughts?

dsteedRTI: Well, the Django docs for [slugify](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#slugify) say that it removes all characters that are not alphanumerics, underscores, or hyphens and replaces spaces with hyphens. So if there aren't any actual alphanumeric characters then its not going to have anything...

dsteedRTI: Although is it very likely that a user will create a project folder without any alphanumeric characters at all?

AstridKery: But if someone does, then the carrot does not work

AstridKery: I mean no, it isn't, but it is a known bug so I wouldn't drop it entirely, since "non-alphanumeric" can eventually happen if someone wanted to use non-english characters or something. 

I would suggest we make an issue to require folder names have some alphanumeric character in the creation forms. I don't think this MR needs that to merge, but I'd like to not lose it. @peridoteagle?

dsteedRTI: Ok, I was just wondering if that was something we needed to have functionality for or if it would be ok to simply require user's to have an alphanumeric character.

dsteedRTI: I added a commit to validate the input field has one alphanumeric. We can use the html field `pattern` on the input field to handle this with a regex string.

AstridKery: Yes this is working now in the project create page. I think we just need to add it to the update page umbrella.html and we should be good

AstridKery (approved):
This looks good to me! Feel free to merge.

peridoteagle (approved):

PR #181, opened 2022-06-09
Npsas main UI improvements

schittarath3-rti:
Some UI improvements off JIRA board


backend/django/core/templates/projects/update/permissions.html
```diff
@@ -81,5 +81,37 @@ <h3>Instructions</h3>
...
+
+// New profile-permission row is added
+$("#permSelect").on("DOMNodeInserted", "tr", function() {
+  dedupPermProfiles();
+})
+
+// Any other change to permission form
+$("#permSelect").on("change", function() {
+  dedupPermProfiles();
+})
```
dsteedRTI, line 115: If this JS is the same in multiple files, you could create a single JS file and import it as a script into both templates.

Perhaps, add a dir in `STATICFILES_DIRS` for scripts and place your script there. Then you can use it a [Django Static File](https://docs.djangoproject.com/en/4.0/howto/static-files/)

frontend/src/components/DataCard/SuggestedLabels.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from "react";
+import { Spinner } from "react-bootstrap";
```
dsteedRTI, line 2: The logic here with React's state management looks great, well done! One small thing I thought of is maybe using React Bootstrap's [Placeholders](https://react-bootstrap.github.io/components/placeholder/) instead of a spinner might look better? This way there's no dramatic content shift and the user can anticipate what the area will look like. What do you think would look better, as this isn't necessarily a change but a discussion!

frontend/src/components/Smart/index.jsx
```diff
@@ -102,15 +102,15 @@ class Smart extends React.Component {
...
+                <Tab eventKey={5} transition={false} title={<div><span className="glyphicon glyphicon-trash" aria-hidden="true"></span> Discarded Data</div>} className="full card">
                     <div className="cardContent">
                         <RecycleBinContainer />
                     </div>
                 </Tab>
             );
         } else {
             adminTabRecycle = (
-                <Tab eventKey={5} transition={false} title={<span className="glyphicon glyphicon-trash" aria-hidden="true"></span>} className="full card">
+                <Tab eventKey={5} transition={false} title={<div><span className="glyphicon glyphicon-trash" aria-hidden="true"></span> Discarded Data</div>} className="full card">
```
dsteedRTI, line 113: I think the `span` could use some margin/padding on the left to give it some space from the trash can icon.

frontend/src/components/History/index.jsx
```diff
@@ -160,8 +184,30 @@ class History extends React.Component {
...
 
+        //Confirmation button
+        let confirm_message = (
+            <Modal
+                centered
+                dialogClassName="history-modal"
+                show={this.state.showConfirm} 
+                onHide={this.toggleConfirm} animation={false}>
+                <Modal.Header closeButton>
+                    <Modal.Title>WARNING</Modal.Title>
```
dsteedRTI, line 195: Instead of "WARNING" in all caps, perhaps let's just do "Confirm" or "Confirmation". Thoughts?

dsteedRTI (changes requested):
Great job on this @schittarath3-rti! The React/frontend logic for these changes looks good and great job working with the Django form for the user permissions dropdown. I added one comment about refactoring some of your JS code into a single script file, and then a couple UI thoughts. Tagging @peridoteagle or @AstridKery if they also have any thoughts on those - particularly the placeholder vs. spinner idea as well as the text for the modal in the history table.

All in all, great job!

frontend/src/components/DataCard/SuggestedLabels.jsx
schittarath3-rti: Ooo I like the look of it. I'll try to add it.

schittarath3-rti: Actually it seems like Placeholder is not in the version of React Bootstrap in the frontend container. We are in version 1.x and Placeholder is in 2.x.

schittarath3-rti:
@dsteedRTI What do you think about these changes? I couldn't get the placeholder to work because (correct me if I'm wrong) it has no been implemented yet.

frontend/src/components/DataCard/SuggestedLabels.jsx
dsteedRTI: hmm, @AstridKery do you think we could/should upgrade React Bootstrap? Or just stick with what we have...?

dsteedRTI:
@schittarath3-rti These look good. Let's talk with the team about the react-bootstrap discussion then this looks ready to me. Well done!

backend/django/core/templates/projects/create/create_wizard_permissions.html
```diff
@@ -36,7 +36,7 @@ <h3>Instructions</h3>
             <li class="list-group-item">Each row must be completely filled in with both a profile and permission.</li>
             <li class="list-group-item">You can update permissions after creating a project</li>
           </ul>
-          <table class="table table-striped">
+          <table class="table table-striped" id="permSelect">
```
AstridKery, line 39: So I was looking at adding permissions to the three user project. I like how the two with permissions are greyed out! one thing I did see though is that we're not catching one user error correctly, and that is the following behavior:

Screen Shot 2022-06-16 at 3 52 26 PM

Someone selects something from the second dropdown but not the first, then we get this error

Screen Shot 2022-06-16 at 3 54 46 PM

This should be simple enough to check for in the form validation section (I believe this would be in the ProjectUpdatePermissions class in frontend.py). Feel free to slack me with any questions!

frontend/src/components/DataCard/SuggestedLabels.jsx
AstridKery: Personally though, I like the spinner :)

AstridKery (changes requested):
Found one form validation check we'll want to add before merging but this looks great!

frontend/src/components/DataCard/SuggestedLabels.jsx
dsteedRTI: Ok sweet, sounds good with me then.

schittarath3-rti:
@AstridKery I fixed that error page. 

One thing I encountered is that when you deselect a profile and select it in a different field. The formset fails to validate.

So using your screenshot:
image

If someone deselects user1 and selects it in the second row and vice versa for test_user, it leads to the error response page.

I'm not sure if we should fix it in this PR or merge this PR and fix it in another PR. It seems like we would need to override the is_valid() function for the permission forms.

AstridKery:
> @AstridKery I fixed that error page.
> 
> One thing I encountered is that when you deselect a profile and select it in a different field. The formset fails to validate.
> 
> So using your screenshot: image
> 
> If someone deselects user1 and selects it in the second row and vice versa for test_user, it leads to the error response page. The actual error message is {'__all__': ['Project permissions with this Profile and Project already exists.']}
> 
> I'm not sure if we should fix it in this PR or merge this PR and fix it in another PR. It seems like we would need to override the is_valid() function for the permission forms.

Interesting! So when I swap the ordering of the two users I'm actually just getting no error but it also won't let me submit. It does look like is_valid is false though. Let's discuss Monday. If it's a simple fix, then it can go here, but if it's a longer haul then it can go in it's own issue.

schittarath3-rti:
> 

Oh my bad. It does not display the error page anymore.

AstridKery (approved):
Looks good to me. The issue with swapping users can be moved to a separate task. @dsteedRTI if you are happy with everything then we can merge

dsteedRTI (approved):

PR #183, opened 2022-06-14
Backup and Add Warning

peridoteagle:
This PR describes the cronjob used for automated backups and adds a warning to the scripts that destroy the DB


dsteedRTI (approved):
These look good to me. Great idea adding a warning to the scripts to avoid what happened last time, LoL!

PR #184, opened 2022-06-15
Changes to Django DB interface with frontend

schittarath3-rti:
- Maximum table name and schema char limit: 50 -> 1024
- On input of DB info on project creation, default to DB on data ingest page


backend/django/core/migrations/0068_auto_20220617_1413.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.2.9 on 2022-06-17 14:13
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
AstridKery, line 6: So this change looks a bit different. Is it possible there is another migration file? When I ran the command, I got a migration with the charfield change like below

```
class Migration(migrations.Migration):

    dependencies = [
        ('core', '0068_auto_20220617_1413'),
    ]

    operations = [
        migrations.AlterField(
            model_name='externaldatabase',
            name='export_schema',
            field=models.CharField(max_length=1024, null=True),
        ),
        migrations.AlterField(
            model_name='externaldatabase',
            name='export_table_name',
            field=models.CharField(max_length=1024, null=True),
        ),
    ]
```

schittarath3-rti:
Sorry for the ugly commit graph. I was trying to figure out how to revert the weird extra file, but it turns out there were un-migrated changes before my changes. Can you check these @AstridKery

AstridKery (approved):
Looks good to me! I think there were some things merged into master though so please merge master into this branch and check if you need to merge the migrations. Just do that by running the migrate command again and it will tell you if you need to run the command with the --merge flag.

schittarath3-rti:
@AstridKery Let me know if this is done correct and if its good to merge!

PR #185, opened 2022-06-21
Adds folder dropdown for project creation/update

schittarath3-rti:
Adds the ability to select folders when adding a new project or updating a project's folder.
Folder selection is based on current user's access to other projects and their folders.

Thoughts:
- Changing Umbrella_string to be separate Django Model which Project FKs to. I think this would make it easier to create empty folders and a basic create folder button on the project list page.
- Making the "Other Projects"  folder more obvious as the default folder.

### DIFF
--- a/backend/django/core/views/frontend.py
+++ b/backend/django/core/views/frontend.py
@@ -29,6 +29,7 @@ from core.models import (
     Label,
     MetaDataField,
     Project,
+    ProjectPermissions,
     TrainingSet,
 )
 from core.templatetags import project_extras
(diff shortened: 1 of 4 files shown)


dsteedRTI:
Created a form accessible from the project listing page to "create a new folder." Allows a user to enter a new folder name and select a project from their existing projects that they either are the creator of or have admin level permissions.

schittarath3-rti:
If the user has no changeable projects on the New Folder page, it leads to this error.
Screen Shot 2022-06-23 at 10 17 23 AM
Screen Shot 2022-06-23 at 10 16 09 AM
Maybe if the project list is empty, change it to a text like "no changeable projects"

schittarath3-rti:
Also should we allow them to remove from folders (and into "Other Projects") by leaving it blank?

schittarath3-rti:
That makes sense, forgot that users can reassign on that page. Everything looks good.

dsteedRTI (approved):

backend/django/core/views/frontend.py
```diff
@@ -200,6 +192,35 @@ def get_form_prefix(self, step=None, form=None):
 
         return prefix
 
+    def get_context_data(self, form, step=None, **kwargs):
+        context = super(ProjectCreateWizard, self).get_context_data(form=form, **kwargs)
+
+        if self.steps.current == "project":
+            user = self.request.user.profile
+
+            if str(user) == "root":
```
AstridKery, line 201: Quick question: why does this user get their own case? I think the root user is just a sample user to test SMART with.

AstridKery:
Looks good to me! Feel free to merge

PR #188, opened 2022-06-27
improve history table

dsteedRTI:
Adds a searchable column to this history table indicating which user labeled the data.

Allows project creators and admins to view all history.


dsteedRTI:
@AstridKery one question I had (and perhaps this is for Emily or the NPSAS team instead) was about who the recorded user for a DataLabel should be? I noticed that if a project creator/admin changes a DataLabel in the History table, the original user who coded the string is still kept as the coder. Does this make sense? Or should this be something we ask the rest of the team about potentially changing?

AstridKery:
> @AstridKery one question I had (and perhaps this is for Emily or the NPSAS team instead) was about who the recorded user for a DataLabel should be? I noticed that if a project creator/admin changes a DataLabel in the History table, the original user who coded the string is still kept as the coder. Does this make sense? Or should this be something we ask the rest of the team about potentially changing?

Hmmmm.... I think it might make more sense for the administrator to be the new coder for the label. The fact that the other profile originally labeled this item should be in the label change log.

AstridKery:
@dsteedRTI what's left for this MR?

dsteedRTI:
@AstridKery I'm going to add that change to record the admin as the user when they update a card, then I'll ping you again and this should be ready.

dsteedRTI:
@AstridKery the history table now records the latest user who modified a DataLabel. ^

AstridKery (approved):
One thing I'll note is that it still counts as a change if you choose the same label it had before, but honestly I don't think that's a big deal. Looks good to me!

PR #190, opened 2022-06-28
Npsas remove uploaded file

schittarath3-rti:
Adds ability to remove codebook from Project Creation page and Update Codebook page
Adds ability to remove labels file from the Project Creation page
- Removes all label fields
- Resets label form to the way it was before the file was uploaded


dsteedRTI:
Also, it looks like you accidentally added `style="text-align-last: center;"` to both inputs, not sure if we want that? It's looking like this on my form, which I don't think looks right:
Screen Shot 2022-06-30 at 3 08 35 PM

schittarath3-rti:
@dsteedRTI Did you mean one comment? I cannot see the other comment.

And to answer the last question, good catch. I'll change the update codebook one back.

dsteedRTI (approved):

dsteedRTI:
Added two things:
- CSS to make our `<button>` tags from now on also look like inline `<p>` tags we have in other areas
- `event.preventDefault()` to the buttons. When you click on a button within a form, they will submit the form. So if you clicked the remove file button it would also attempt to submit the form... to stop this, you can add the button's `event` to the function and the prevent it from submitting by using `event.preventDefault()`

PR #193, opened 2022-06-30
Forbid duplicate data with labels from being ingested

schittarath3-rti:
This throws a form error is duplicate data with old labels in ingested through the update data pipeline. Does not affect project creation data upload.

Error shows which rows in the uploaded data contains this problem

### DIFF
--- a/backend/django/core/forms.py
+++ b/backend/django/core/forms.py
@@ -499,10 +499,14 @@ class DataUpdateWizardForm(forms.Form):
 
             # Throw error containing rows which are going to overwritten
             if not overwrite_data.empty:
-                error_str = "The following rows are already in the database and the labels will be overwritten:\n\n"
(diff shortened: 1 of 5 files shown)


schittarath3-rti:
I think the part where it shows which rows are problem-causing is not user-friendly. 

Currently, it shows it as a csv, but pandas has a to_html function for dataframes which could look better. It does not work well with django form errors though because it shows it has the string rather than html.

schittarath3-rti:
Current error message:
image

schittarath3-rti:
Also what do you guys think about limiting the number of rows that is displayed? Currently, it shows all rows.

AstridKery:
I'd say list the first 5 or 10, then just say the number of remaining differences. The error message is a bit confusing. Are you saying the data is already there and labeled, or it's there at all?

schittarath3-rti:
Should mean that the data is there, either unlabelled or labelled.

PR #194, opened 2022-07-01
add option to skip queue without adjudicating (for later)

dsteedRTI:
Adds a button to just "Skip" a card. This does almost everything the same as our current "Adjudicate Button" except for adding that card to the Admin queue.

Renames the "Skipped" tab to "Requires Adjudication"

### DIFF
--- a/backend/django/core/views/api_annotate.py
+++ b/backend/django/core/views/api_annotate.py
@@ -123,6 +123,28 @@ def label_distribution_inverted(request, project_pk):
 
 @api_view(["POST"])
 @permission_classes((IsCoder,))
+def unassign_data(request, data_pk):
+    """Take a datum that is in the assigneddata queue for that user and remove it from
+    the assignedData queue.
+
+    Args:
+        request: The POST request
+        data_pk: Primary key of the data
+    Returns:
+        {}
+    """
+    data = Data.objects.get(pk=data_pk)
+    profile = request.user.profile
+    response = {}
+
+    assignment = AssignedData.objects.get(data=data, profile=profile)
+    assignment.delete()
+
+    return Response(response)
+
+
+@api_view(["POST"])
+@permission_classes((IsCoder,))
 def skip_data(request, data_pk):
     """Take a datum that is in the assigneddata queue for that user and place it in the
     admin queue. Remove it from the assignedData queue.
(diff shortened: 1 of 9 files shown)


frontend/src/components/DataCard/CardData.jsx
```diff
@@ -29,7 +30,7 @@ function drawSkipButton(card, onSkip) {
 
     const handleSkip = (event) => {
         event.preventDefault();
-        onSkip(card, message);
+        onSkip(card, true, message);
```
AstridKery, line 33: Looks like this needs to be updated in history as well. Right now when I send to administrator the reason is set as "true" or "false" instead of my reason.

backend/django/core/views/api_annotate.py
```diff
@@ -159,8 +159,9 @@ def skip_data(request, data_pk):
             process_irr_label(data, None)
     else:
         # the data is not IRR so treat it as normal
-        move_skipped_to_admin_queue(data, profile, project)
-        createUnresolvedAdjudicateMessage(project, data, request.data["message"])
+        if request.data["addToAdminQueue"]:
```
AstridKery, line 162: why is this added here? Doe that mean if data is skipped using the skip button and is IRR then it is send through the administrator skip system normally? I'm not sure I agree with that, since I thought the skip button mean "I want to look at this later" which is not the same as "I think this is uncodable"

AstridKery:
Hey @dsteedRTI ! Sorry for the wait. I think we should discuss this Monday, having some concerns about the workflow with some of the cases of using the skip button.

frontend/src/components/DataCard/CardData.jsx
```diff
@@ -29,7 +30,7 @@ function drawSkipButton(card, onSkip) {
 
     const handleSkip = (event) => {
         event.preventDefault();
-        onSkip(card, message);
```
AstridKery, line 32: Actually though, right now, the skip button exists in the history table, but what exactly does that mean? I don't think it should really render there.

backend/django/core/views/api_annotate.py
```diff
@@ -121,6 +121,28 @@ def label_distribution_inverted(request, project_pk):
...
+    Returns:
+        {}
+    """
+    data = Data.objects.get(pk=data_pk)
+    profile = request.user.profile
+    response = {}
+
+    assignment = AssignedData.objects.get(data=data, profile=profile)
+    assignment.delete()
+
```
AstridKery, line 142: Need to add the check_and_trigger_model call which is also at the bottom of skip_data. This will refill their deck if they've skipped their last card.

dsteedRTI: Good catch!

AstridKery: Actually twas a batch catch all along. This function checks if the queue is empty, but of course you're not messing with the queue at all. You actually don't need this. A simple refresh of the page will give them new cards.

peridoteagle:
Is this good to merge?

AstridKery:
I thought we had already merged this. Whoops!

PR #198, opened 2022-07-11
Create project UI fixes

schittarath3-rti:
- Change instructions of data upload to clarify that label field is not mandatory
- Labelled project name field as required
- Changed ordering of tabs
  - Advanced tab is optional
  - Advanced tab is moved to step 7
-  Example dataset link text is entirely highlighted


dsteedRTI:
After discussing with @AstridKery let's keep the accordions closed by default. I pushed some padding to keep it like the other pages in the form. Also put the "Upload File" button for the Codebook page back to being left-aligned after discussing with @peridoteagle

PR #201, opened 2022-07-12
Error message if login fails for db connection

schittarath3-rti:


AstridKery:
Hey @schittarath3-rti ! I missed when this task was handed out. Which DB errors are we specifically worried about catching here? Is there a Jira issue?

schittarath3-rti:
It catches an incorrect password error, but because there isn't one. It catches a generic account verification error. EWDCODE-119

backend/django/core/utils/utils_external_db.py
```diff
@@ -25,6 +26,24 @@ def get_connection(db_type, connection_dict):
         raise ValueError(f"ERROR: database type {db_type} is invalid")
 
 
+def test_login(db_type, engine_database):
+    """Attempts to login with given connection."""
+    try:
+        # Perform a no-op
+        pd.read_sql("NULL", con=engine_database)
```
AstridKery, line 33: Getting this error when I put in the correct password
Screen Shot 2022-07-18 at 3 48 57 PM

AstridKery (approved):
This is working for me now. Good work!

PR #203, opened 2022-07-20
Npsas UI Improvements: 2nd Edition

schittarath3-rti:
- Adjusted quartiles on Time To Label graph so that Q0, Q4 show up and all values have units
- Profiles are searchable on permissions page
- Confirmation modal opens when user wants to remove permission from project


dsteedRTI (approved):
Looks great, well done! Added some styles to the select to try and mimic the `<select>` tag's styling like we discussed earlier.

PR #206, opened 2022-08-01
patched data unassigned after being assigned

schittarath3-rti:
I referenced this in #205, but this issue prevents data from being labelled due to data being unassigned after getting assigned. Now, I know the issue appears in the main branch.

### DIFF
--- a/backend/django/core/utils/utils_annotate.py
+++ b/backend/django/core/utils/utils_annotate.py
@@ -26,7 +26,7 @@ from core.utils.utils_redis import (
 
 def leave_coding_page(profile, project):
     """unnasign data and remove any admin locks for a specific user and project."""
-    assigned_data = AssignedData.objects.filter(profile=profile)
+    assigned_data = AssignedData.objects.filter(profile=profile, data__project=project)
 
     for assignment in assigned_data:
         unassign_datum(assignment.data, profile)


schittarath3-rti:
Example API request order
image

AstridKery:
Seeing this line from the referenced MR. Is this what you mean? And if so, would you mind clarifying what the fix does? Thanks :) 

"In the case where get_card_deck is called before enter_coding_page, data is assigned, but then unassigned in enter_coding_page. There is a commit that fixes it here, but I am not sure if that is a problem in npsas_main"

schittarath3-rti:
Correct.

So what's happening is that `get_card_deck` is being called before `enter_coding_page`. `get_card_deck `assigns the data to the coder, and `enter_coding_page` unassigns all data for all projects. In the old version, `assigned_data` selects _all_ data for the profile, not specifically the projects iterated in `enter_coding_page`

_api_python.py>enter_coding_page_
```python
profile_projects = [p for p in profile_projects if p != project]
    for project in profile_projects:
        leave_coding_page(profile, project)
```

util_annotate.py>leave_coding_page
```python
assigned_data = AssignedData.objects.filter(profile=profile)
```

We need to add `data__project = project` so it selects only the assigned data for that project.

This problem does not occur if `get_card_deck` is called after `enter_coding_page` because the data for the project is just assigned afterwards.

AstridKery (approved):
Got it, approved, and good catch!

PR #207, opened 2022-08-04
Manual unassign coders

schittarath3-rti:
Adds the ability to manually unassign data from coders through the Admin dashboard.

For future reference, the manual unassign problem ([EWDCODE-197](https://jira.rti.org/browse/EWDCODE-197)) occurs when the changes in the database (ie removing AssignedData entires) is not reflected in the redis queue that is responsible for distributing data to coders.

### DIFF
--- a/backend/django/core/templates/projects/admin/admin.html
+++ b/backend/django/core/templates/projects/admin/admin.html
@@ -24,6 +24,9 @@
           <a data-toggle="tab" href="#irr_tab" class="nav-link">IRR</a>
         </li>
         {% endif %}
+        <li class="nav-item" id="unassign_tab_navitem">
+          <a data-toggle="tab" href="#unassign_tab" class="nav-link">Unassign Coder</a>
+        </li>
       </ul>
     </div>
     <div class="tab-content">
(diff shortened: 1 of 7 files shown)


peridoteagle:
@AstridKery this looks ok to me, but wanted to run it by you first

AstridKery:
Seeing a few issues here:

* if a coder has their stuff unnasigned, and they try to label a card, SMART throws an internal error and the submit doesn't work but there's nothing to tell the coder why the button isn't working. To them SMART just seems frozen. We need to properly handle this use case in the coder actions instead of just letting SMART error. 
* Need to add admin permission header to the api call (one line change)

AstridKery:
@peridoteagle I've made some updates to address my comments. Would you mind taking a look? Note that it helps to be signed in as both an admin and a non-admin user on different screens so you can see the impact.

AstridKery:
@peridoteagle Question: this is an admin action that I don't think is blocked by the adminaccess thing, so in theory multiple admin could remove coders at once. Do we care?

AstridKery:
@peridoteagle I tested this out and actually the unassign_coders function doesn't error if a second admin calls it before refreshing the page. The function will just look for assignments and not remove anything. So I think this is ready for your review.

peridoteagle (approved):

PR #208, opened 2022-08-09
add profile and timestamp to data export

AstridKery:
This change adds the profile and timestamp to the exports. 

This should be very simple for new projects, but for existing ones, we'll need to think of a workaround. One option is to change the export table name and re-export all items. 

Another would be to add the Profile and Timestamp columns manually and then just have it add those fields for the newly labeled items. Unless this MR should append the columns if they are not there?

### DIFF
--- a/backend/django/core/forms.py
+++ b/backend/django/core/forms.py
@@ -10,11 +10,12 @@ from pandas.errors import ParserError
 
 from core.utils.util import md5_hash
 from core.utils.utils_external_db import (
+    check_if_schema_exists,
+    check_if_table_exists,
     get_connection,
     get_full_table,
     test_connection,
     test_login,
-    test_schema_exists,
 )
 
 from .models import ExternalDatabase, Label, Project, ProjectPermissions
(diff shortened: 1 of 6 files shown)


AstridKery:
UPDATE: now SMART checks if the export table exists and will tell the user to pick a different table when they set it up

peridoteagle (approved):
I tested this on staging with the test database and it worked, adding the user and timestamp. This was an export to an existing table that did not previously have user and timestamp columns, and I noted that it wrote over the table.

Is this different than current export behavior and if so, do you have language that I should use to let Donna know about the change?

Perhaps also letting the user know that an export table should JUST be used for database exports because it will be written over. Let me know what you think.

AstridKery:
Hi @peridoteagle so yes this is different from before, where we would just append new ID's to the export table. We discussed how that method was failing to update items which were changed in the history table. I just updated the warning message to include that, although this will not help with projects that already have export set up.

peridoteagle:
That makes sense. Does the following look like an accurate description of what I send to Donna?

_"One critical change: exporting data to a SQL database will now export the entire history in SMART and rewrite whatever exists in the export table in the database (rather than appending new rows). The main reason for this is that it will capture changes to the history made in SMART. The secondary reason for this is that it will add user and date/time for items that are already labeled. The impact of this is that no manual changes or additions should be made to the export table in the SQL database as these will be overwritten anytime the data is exported. The export table in the SQL database should be for SMART exports only"_

AstridKery:
@peridoteagle yes that looks correct to me!

peridoteagle:
Great! I'm going to go ahead and merge this and send Donna the language above

PR #211, opened 2022-08-18
pin django higher for security updates

AstridKery:


AstridKery:
@dsteedRTI we should probably resolve this one when we have a few hours

AstridKery:
resolved in another MR

PR #212, opened 2022-08-23
make queue refill properly when assignedData is removed from database

AstridKery:
This also fixes a potential bug where projects without classifiers but with IRR might have only been refilling the regular queue and not the IRR queue when they ran out of things to label.


peridoteagle (approved):
Looks good to me. Nice work!

PR #213, opened 2022-08-25
fix a bunch of broken tests so they pass

AstridKery:
Not urgent: I just noticed these were failing and wanted to get all the tests passing again


AstridKery:
@rjruizes this is ready for review

PR #215, opened 2022-08-28
Change timezone to EST on export

AstridKery:
This MR just changes the timezone of the export to Eastern time instead of UTC. 

NOTE: This is one of those changes that works for this project but will probably get confusing for someone else looking to pick up SMART in the future. Not sure if there's somewhere we could make a note of this change in the README or docs?


epugh:
You could add a system property/configuration variable that specifies the output time zone format?

dsteedRTI:
I like this way a lot, looks great

PR #223, opened 2022-10-03
create timestamp now default for datalabels

dsteedRTI:
Adds a default value for timestamps with `now` so that exports are not breaking due to null timestamp values.


dsteedRTI:
This branch does two things:

1. Specifies a default of `timezone.now` for all data labels created through the csv upload. Django threw errors if there was a null or none value for timestamp on creation, so this goes ahead and gets the current timestamp for each data label as its created.
2. Adds a default value to the `timestamp` column. For all timestamps in the database that were previously created through the csv upload option, they currently have null timestamps. This will ensure that they have an actual timestamp value so that the export function will now work for every project.

PR #227, opened 2022-10-14
More minor UI fixes

AstridKery:
This branch addresses the following:
* fixing the IRR heatmap to be readable (only shows labels which have been used, gives more space for the label names)
* Replace "Ingest" with "Input" in directions
* Show "Other groups" instead of a blank when adding projects to groups
* make label icon arrows instead of +/-


dsteedRTI:
Screen Shot 2022-10-17 at 4 25 04 PM

I'm getting this error on the IRR tab for the 3 user default project. Loading the heat map on a separate branch and I'm not running into this error? Not sure if it's only on the default 3 user project or on all projects with IRR...

PR #228, opened 2022-10-24
Add descriptions to labeled data

dsteedRTI:
I added a conditional so that if every label in the entire project does not have a description, no description column will show up on the exported csv/table and clog it up with empty values. Thoughts?


AstridKery (approved):
Tested the affects of this on history, labeled data, and database exports and it seems to work fine

PR #232, opened 2022-11-10
Export cron jobs

AstridKery:
This adds automated exports to the codebase. It should work the same way as the imports. Let me know what you think!


peridoteagle (approved):
Tested on staging and this works. Will merge to production

PR #241, opened 2022-12-14
History table updates

dsteedRTI:
(I know we're going to rename the History table to something else... however; still going to refer to it as the History table in this PR until we settle on a new name.)

- adds dynamic columns to table for metadata columns
- improves CSS and React Table options to optimize for more columns
- Adds debounced inputs for all metadata values on the table:
This allows us to automatically save any edits to the table to our database *once the user finishes typing in order to avoid any unnecessary API calls/database updates

### DIFF
--- a/backend/django/core/views/api_annotate.py
+++ b/backend/django/core/views/api_annotate.py
@@ -956,9 +956,7 @@ def modify_metadata_value(request, metadata_pk):
     Returns:
         {}
     """
-    print(metadata_pk)
     metadata = MetaData.objects.filter(pk=metadata_pk).first()
     metadata.value = request.data["value"]
-    print(request.data["value"])
     metadata.save()
     return Response({})
--- a/backend/django/core/views/api_annotate.py
+++ b/backend/django/core/views/api_annotate.py
@@ -22,6 +22,7 @@ from core.models import (
     Label,
     LabelChangeLog,
     LabelEmbeddings,
+    MetaData,
     Profile,
     Project,
     Queue,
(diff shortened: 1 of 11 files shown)


backend/django/core/views/api_annotate.py
```diff
@@ -926,3 +942,23 @@ def get_label_history(request, project_pk):
         results.append(temp_dict)
 
     return Response({"data": results})
+
+
+@api_view(["POST"])
+# @permission_classes((IsCoder,))
+def modify_metadata_value(request, metadata_pk):
```
AstridKery, line 949: So the only issue I'm seeing with this is that you have to click away from the history tab and back for the change to show up when you maximize the item. 

Not really a problem though I think? Just something the users have to be aware of. In general I'd add a line to the instructions on the top of the page. Something like "In this table you may edit metadata fields. Click on the value in the column and row where you want to change the data and it will open as a text box."

backend/django/core/views/api_annotate.py
```diff
@@ -926,3 +942,23 @@ def get_label_history(request, project_pk):
...
+    """Take a single datum with a label and change the label in the DataLabel table.
+
+    Args:
+        request: The POST request
+        metadata_pk: Primary key of the metadata
+        value: New value for metadata field
+    Returns:
+        {}
+    """
+    print(metadata_pk)
```
AstridKery, line 959: are these prints needed?

frontend/src/components/History/EditableMetadataCell.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { useState } from "react";
+import { DebounceInput } from "react-debounce-input";
+
+export default function HistoryMetadataCell({ id, value, modifyMetadataValue }) {
```
AstridKery, line 4: Would it be possible to add tooltips to the headers so when you hover on a column it will give the full title? The metadata field names get cut off at about 11 characters which could make it hard to distinguish them.

dsteedRTI: Added for all metadata columns

frontend/src/components/CodebookLabelMenu/index.jsx
```diff
@@ -97,13 +97,14 @@ class CodebookLabelMenu extends React.Component {
...
-                    </ButtonGroup>
+                    {CODEBOOK_URL != "" ? (
+                        <ButtonGroup className="pull-left">
+                            {label_button}
+                            {codebook_button}
+                        </ButtonGroup>
+                    ) : label_button}
                 </div>
                 {this.getLabels(labels, this.state.labels_open)}
-                {codebook_module}
```
AstridKery, line 106: I'm not getting the codebook to pop up when I click the button. I'm on Chrome. Have you had this issue?

dsteedRTI: I'm not even sure how/why the lines for the codebook module got removed but I've added them back in and things seem to be working now.

PR #242, opened 2022-12-29
Add labeled data verification

AstridKery:
This branch adds the concept of unverified labels. These are defined as any data labels created outside of SMART and imported/ingested into the project. The main concept behind this update is that sometimes people load old labeled data into projects to avoid having to re-label the same strings, but there might be some desire to verify that old data for some reason (labels may have changed definitions a bit, the older labeling effort had less overhead, etc.)

Users can verify unverified data in the history tab. **NOTE:** as this MR also affects the history table, we will need to reconcile it with the history overhaul MR.

In cases where projects have both verified and unverified data labels they can choose to download all labeled data or just verified labeled data.

In addition the automated export now includes an option to only export verified labeled data.


backend/django/core/templates/projects/detail.html
```diff
@@ -263,10 +266,16 @@ <h5 class="panel-title">
...
           {% if project.has_model %}
-            <a class="btn btn-primary" id="download_model_btn">Download Model and Labeled Data</a>
+            <a class="btn btn-primary" id="download_model_btn" onclick="download_model({{project.pk}},1)">Download Model and All Labeled Data</a>
+            {% if project.unverified_labeled_data_count > 0 %}
+              <a class="btn btn-primary" id="download_model_btn_verified" onclick="download_model({{project.pk}},0)">Download Model and Only Verified Labeled Data</a>
+            {% endif %}
           {% else %}
             {% if project.labeled_data_count > 0 %}
-            <a class="btn btn-primary" id="download_btn">Download Labeled Data</a>
+            <a class="btn btn-primary" id="download_btn" onclick="download_data({{project.pk}},1)">Download Labeled Data</a>
```
dsteedRTI, line 275: The buttons on the project detail page; however, are correctly downloading data.

frontend/src/components/History/index.jsx
```diff
@@ -238,7 +259,7 @@ class History extends React.Component {
                 <CodebookLabelMenuContainer />
                 <ReactTable
                     data={history_data}
-                    columns={[...COLUMNS, ...metadataColumns.map((column, i) => {
+                    columns={[...this.COLUMNS(), ...metadataColumns.map((column, i) => {
```
dsteedRTI, line 262: I don't think this is specific to this branch but leaving a comment here to document it. It looks like attempting to sort the history table based off a metadata column shifts theorizer of the left-hand columns but any metadata-related columns do not reorder.

dsteedRTI (changes requested):
The logic behind verified/unverified data worked perfectly for me on many different project tests! The only major issue I could find is correctly downloading a project's verified or unverified data from the project list page.

In addition, there is something funky with the history table but I'm not sure if that is specifically related to this branch (probably not.)

backend/django/core/templates/projects/list.html
```diff
@@ -72,10 +72,16 @@ <h5 class="panel-title">
...
                         {% if project.has_model %}
-                          <td><a class="download_btn" name="{{ project.pk }}" onclick="download_model({{project.pk}})">Download Model and Labeled Data</a></td>
+                          <td><a class="download_model_btn" name="{{ project.pk }}" onclick="download_model({{project.pk}},1)">Download Model and Labeled Data</a></td>
+                          {% if project.unverified_labeled_data_count > 0  and project.labeled_data_count > project.unverified_labeled_data_count %}
+                            <td><a class="download_btn" id="download_model_btn_verified" onclick="download_model({{project.pk}},0)">Download Model and Only Verified Labeled Data</a></td>
+                          {% endif %}
                         {% else %}
                           {% if project.labeled_data_count > 0 %}
-                          <td><a class="download_btn" name="{{ project.pk }}" onclick="download_data({{project.pk}})">Download Labeled Data</a></td>
+                          <td><a class="download_btn" name="{{ project.pk }}" onclick="download_data({{project.pk}},1)">Download Labeled Data</a></td>
```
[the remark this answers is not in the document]
AstridKery: Thanks for catching that! Turns out I had only updated the code on the default project section of the if/else. Just pushed an update

frontend/src/components/History/index.jsx
AstridKery: I'm not sure! This is really odd. I ran this in dev and it has the same issue.

AstridKery:
@dsteedRTI This MR has been updated with the changes discussed in slack. We have pre-loaded, which only changes if someone edits the existing label, and verified, which starts out false and must be swapped to True in the history table

NOTE: Currently, I have verified being set to True if an admin labels it in the admin table, or it is resolved IRR. I'm going to make a chat in the slack about this though, since it's a bit complicated and I'm not sure if I should change it

dsteedRTI (commented):
Wondering if it would be a good idea to include a count of pre-loaded data on the details page... It might be confusing to see XX number of fully labeled data, but to also see 0 fully verified or fully unverified data?
Screen Shot 2023-03-07 at 5 15 01 PM

frontend/src/components/History/index.jsx
```diff
@@ -72,6 +21,86 @@ class History extends React.Component {
...
+                Header: "Verified",
+                width: 80,
+                accessor: "verified",
+                Cell: props => {
+                    if (props.value == "Yes") {
+                        return (<p>Yes</p>);
+                    } else if (props.value != "No") {
+                        return (<p>{props.value}</p>);
+                    } else {
+                        return (<Button variant="success" value={props.row.id} onClick={this.verifyLabel}>Verify</Button>);
```
dsteedRTI, line 87: Perhaps I don't understand, but when I verify pre-loaded data, I would expect to see if add to the count of fully verified data on the Details page. Its not for me, but is that expected behavior?

AstridKery: It should be! Let me check

AstridKery: Good catch! Yes the logic hand't been fixed after the update. I pushed a change now

frontend/src/components/History/index.jsx
AstridKery: Closing this since we're addressing in a different branch

frontend/src/components/History/index.jsx
dsteedRTI: Yes, things look correct now!

dsteedRTI (approved):

PR #244, opened 2023-03-07
allow toggle for unlabeled data in history table

dsteedRTI:
Two thoughts:

- Depending on how many unlabeled data are in a project, a little worried about how long requests to get all of the data could take... thoughts?
- Noticing some weird/lagging search results for some metadata columns in the History table

### DIFF
--- a/frontend/src/components/History/index.jsx
+++ b/frontend/src/components/History/index.jsx
@@ -67,7 +67,8 @@ class History extends React.Component {
         this.historyChangeLabel = this.historyChangeLabel.bind(this);
         this.state = {
             pageSize : parseInt(localStorage.getItem("pageSize") || "25"),
-            showConfirm : false
+            showConfirm: false,
+            showUnlabeled: false
         };
         this.cardID, this.rowID, this.label;
     }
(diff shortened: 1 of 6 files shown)


frontend/src/components/History/index.jsx
```diff
@@ -178,7 +183,7 @@ class History extends React.Component {
     }
 
     render() {
-        const { history_data } = this.props;
+        const { history_data, unlabeled_data } = this.props;
```
AstridKery, line 186: So I think for the use case we generally see here, it might make more sense to merge these two lists on the backend, sort alphabetically, and then return. It seems like they're going to be comparing similar data that is both labeled and unlabeled side by side.

frontend/src/components/History/index.jsx
```diff
@@ -236,8 +241,17 @@ class History extends React.Component {
...
+                        Toggle the checkbox below to show/hide unlabeled data:
+                    </p>
+                    <label>
+                        <input type="checkbox" onChange={() => this.toggleShowUnlabeled()} />
+                        <span style={{ marginLeft: "4px" }}>Unlabled Data</span>
+                    </label>
+                </div>
                 <ReactTable
-                    data={history_data}
+                    data={this.state.showUnlabeled ? [...history_data, ...unlabeled_data] : history_data}
```
AstridKery, line 254: See above. I think it would make sense to just have unlabeled be an argument to the API call and history data includes it or leaves it out based on that

PR #246, opened 2023-03-17
Small UI improvements

AstridKery:
This branch adds the following:

* description of verification to the details page
* reset filters button to the history table
* error handling for double clicking in the history table (ignores multiple labeling calls in a row)
* makes search case-insensitive
* rename page to batch in dropdown


peridoteagle:
Looks great - nice work @AstridKery!

PR #248, opened 2023-03-27
update metadata

dsteedRTI:
Adds the ability to update metadata values on a Card itself.

- Works from the History Table; however, unsure how to force the History table to refresh... thoughts? EDIT: this has been solved, thanks!


frontend/src/actions/card.js
```diff
@@ -144,3 +144,23 @@ export const passCard = (card, num_cards_left, is_admin, projectID, message) =>
             });
     };
 };
+
+// update a card's metadata
+export const modifyMetadataValues = (dataPk, metadatas, projectPk) => {
```
AstridKery, line 149: So I was clicking through this in the annotate data tab and discovered the following behavior:

* I clicked "edit" on an item, changed something, and clicked "save". The metadata item changed.
* I clicked "edit" again on the same item, but this time just selected "cancel". **This changed the value back to what it was before my initial edit.**
* I clicked "edit" again and tried to change the metadata item back. The following error was thrown
```
backend              | Internal Server Error: /api/modify_metadata_values/7245/
backend              | Traceback (most recent call last):
backend              |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
backend              |     response = get_response(request)
backend              |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
backend              |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
backend              |   File "/usr/local/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
backend              |     return view_func(*args, **kwargs)
backend              |   File "/usr/local/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
backend              |     return self.dispatch(request, *args, **kwargs)
backend              |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 509, in dispatch
backend              |     response = self.handle_exception(exc)
backend              |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 469, in handle_exception
backend              |     self.raise_uncaught_exception(exc)
backend              |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
backend              |     raise exc
backend              |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
backend              |     response = handler(request, *args, **kwargs)
backend              |   File "/usr/local/lib/python3.8/site-packages/rest_framework/decorators.py", line 50, in handler
backend              |     return func(*args, **kwargs)
backend              |   File "/code/core/views/api_annotate.py", line 1093, in modify_metadata_values
backend              |     metadata = MetaData.objects.get(data_id=data.id, value=m["previous"])
backend              |   File "/usr/local/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
backend              |     return getattr(self.get_queryset(), name)(*args, **kwargs)
backend              |   File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 435, in get
backend              |     raise self.model.DoesNotExist(
backend              | core.models.MetaData.DoesNotExist: MetaData matching query does not exist.
```

I suspect the issue here is with the cancel button. It's dropping all changes in what's displayed to the user, including the ones previously saved. The backend has changed though, so it can't find the unchanged value.

I would recommend having the form do something like store the last saved values for cancel to reset to, or else keep a list of changes made since the last call to save.

frontend/src/actions/card.js
```diff
@@ -144,3 +144,23 @@ export const passCard = (card, num_cards_left, is_admin, projectID, message) =>
...
+                if (response.ok) {
+                    return response.json();
+                } else {
+                    const error = new Error(response.statusText);
+                    error.response = response;
+                    throw error;
+                }
+            })
+            .then(() => {
+                dispatch(getHistory(projectPk));
```
AstridKery, line 163: I'd recommend adding `dispatch(getUnlabeled(projectID));` here, so the skew table updates. If you change something then close and re-open the card it doesn't keep the change.

AstridKery (approved):
Looks great!

PR #250, opened 2023-04-05
reset batch page when filtering so you don't get stuck

AstridKery:
Quick bug fix for issue found in filtering. Basically if you go to a high batch and then filter, the batch number doesn't reset and the history table is then trying to return the fifth page of results that may not go that far. Thus you typically get nothing and the batch dropdown dissapears.


AstridKery:
NOTE: this doesn't fix another behavior documented in Jira (issue 248) however that one is going to require a bit more re-organizing. This MR is just meant to be a quick fix.

PR #253, opened 2023-04-14
History table and admin chart changes requested

AstridKery:
This Branch makes the following changes

* Change "Old Label" to "Current Label" in history table
* Re-order history table columns. Move label user, time, and validation fields to the right of metadata
* Re-name "User" and "Date/Time" to "Labeled By" and "Labeled Date/Time"  to make it easier to understand without them being right next to the label column
* GOAL: Give the strings in the history table columns more space before cutting off **NOTE: Not sure what to do here. Technically it would make sense to use the lengths of the items in the metadata to define the width, but this might be more complicated than we need it to be since people can just resize. @peridoteagle thoughts?**
* Move dropdown and label buttons to the left side of the card so the user doesn’t have to scroll right
* **NOTE: I'd also like to move the skil/adjudicate buttons as well as the edit button for the metadata to be on the left, but I'm having trouble tracking down the right CSS to change @dsteedRTI do you remember?**
* Don't show the pre-loaded data in the admin table label distribution, since it shows a huge column assigned to the project creator. **Question for @peridoteagle : should we add a note telling users pre-loaded data isn't included, or change the headers for the charts?**


AstridKery:
After talking about this in the meeting, I have the following final updates:

> This Branch makes the following changes
> 
> * Change "Old Label" to "Current Label" in history table
> * Re-order history table columns. Move label user, time, and validation fields to the right of metadata
> * Re-name "User" and "Date/Time" to "Labeled By" and "Labeled Date/Time"  to make it easier to understand without them being right next to the label column
> * Move dropdown and label buttons to the left side of the card so the user doesn’t have to scroll right
> * Don't show the pre-loaded data in the admin tab, since it shows a huge column assigned to the project creator and rename labeled data table to "Data Labeled in SMART"

I'm going to make another issue for hiding history table columns, wrapping text, and adding filters for label and verification fields

peridoteagle (approved):
I tested this out on staging and it looks great.

Re column width: I personally think the column resizing is sufficient to address the needs of reading the text. Hopefully this functionality is intuitive to folks, but could consider adding a note or instruction i the future if users complain.

Re admin chart: I agree with excluding pre-labeled data. I think it's intuitive given that the Y-axis label is "Number of Data Annotated", but could be persuaded to have a note added that says something like, "Pre-loaded labeled data not included" if users ask questions

PR #255, opened 2023-05-19
fix the database export so the verified fields are in the correct format

AstridKery:
This MR should fix the database errors from exporting data with the verified fields.


peridoteagle (approved):
Looks good - testing out on staging now

peridoteagle:
Tested on staging with a confirmed export to the test database with no errors. Pushing to production.

PR #259, opened 2023-06-22
Bug Fix for Zach Regarding Error When New Line Break in Excel

peridoteagle:
This bug caused SMART to break when importing an Excel file with a new line break. This change forces new line breaks to be spaces.


AstridKery (approved):
Looks good. This is the only place I see read_excel, so I think this should be the extent of it.

PR #260, opened 2023-07-05
ML upgrades

dsteedRTI:
MVP for adding unique, trainable models project-specific on SMART.

## Docker

Created a new `ml` container to do all ML-related processes in.

This container has several volumes, `/csv` and `/models` which store the `.csv` files used to train models and the actual models respectively.

## ML

Removed all embeddings/`sentencetransformers` code from the `backend` and moved it to the new `ml` container. These processes are now available on a [FastAPI](https://fastapi.tiangolo.com/) REST API that will interact with the `backend`.

To train models, abstracted our [csv-to-embeddings-model](https://github.com/RTIInternational/csv-to-embeddings-model) process to work with an uploaded `csv` sent to `ml`'s FastAPI.

When SMART now goes to encode an embedding/compute similarity comparisons, `backend` sends a request to `ml` with the desired string(s) to encode. Within `ml`, it first looks to see if a model is already trained under `/models`. If one exists, it uses that specifically trained model to compute the embeddings; if not is uses our default SMART model.

## Backend Updates

Updated all existing backend ML-related processes to integrate with the new `ml` container.

Created new `MlModel` in Django models.

Added a new template accessible through a project's Update page, "Update ML". On this page an admin can upload a `.csv` file with text pairs and train a new model to be used on that project. This process involves:

- Update/create a `MlModel` and set its status to "Training"
- Call `ml`'s FastAPI `/train/` endpoint to kick off the model training process, forwarding the uploaded `.csv` file.
- Train the actual model in the `ml` container.
- Update the `MlModel` and set its status to "Trained".
- Update existing `LabelEmbeddings` to be re-computed with the newly trained model

## Next Steps

Couldn't get persistent storage for both the `/csv` or `/models`/ volumes in `ml`... probably just a Docker thing I'm messing up so if somebody can get a second set of eyes on that and see what's not configured properly that would be very helpful!

Once SMART is moved to an entirely React-based frontend, we can make a reusable hook with React Query for training the model. Unfortunately, React only works on the Annotate tab right now and so the form/logic for sending the request to `ml` had to be done through Django at this time.


hardingalexh:
@dsteedRTI adding this as a comment because individual line-level comments could get confusing here. Regarding your docker persistence issue, you'll need to name the volumes. What's a little wacky is that in the current pattern with extending a common docker-compose file, you're naming those volumes in the `dev` and `prod` docker-compose files, but referencing them in the `common` file (so you'll need to declare the volumes in both the `dev` and `prod` files).

For consistency's sake, you can do the following:

* In the `volumes` section of `/envs/dev/docker-compose.yml`, add your named volumes like so:
```yml
volumes:
  smart_pgdata:
    external: true
    name: vol_smart_pgdata
  smart_data:
    external: true
    name: vol_smart_data
  ml_csv:
  ml_models:
```

* In your `envs/docker-common.yml` file, use those named volumes in your service definition for the `ml` service:
```yml
ml:
    container_name: ml
    volumes:
      - ../ml:/code
      - ml_csv:/code/csv
      - ml_models:/code/models
    ports:
      - "${EXTERNAL_ML_PORT:-8001}:8001"
```

Those containers should then be persistent. 


---
One other note I have is that in `docker-common.yml`, the application is configured to bind-mount from host a number of things:
```yml
backend:
    container_name: backend
    volumes:
      - ../backend/django:/code ## bind mount
      - ../backend/docker:/code/docker ## bind mount
      - ../frontend:/code/frontend ## bind mount
      - smart_data:/data/
    ports:
      - "${EXTERNAL_BACKEND_PORT:-8000}:8000"
    command: ./runserver.sh

  smart_frontend:
    container_name: smart_frontend
    volumes:
      - ../frontend/:/code ## bind mount
      - /code/node_modules
```

This is typically something that you would do in `dev` to enable hot reloading, but in prod you would `COPY` those files in your `Dockerfile`. This isn't a huge issue, but it does mean that your production application is bind mounting some directories on the host which can cause future problems with directory permissions and/or performance. I'd recommend at some point moving the bind mounts to the dev dockerfile and ensuring that everything is copied when building for prod.

dsteedRTI:
@hardingalexh I can confirm implementing those Docker configurations fixed persistent storage on dev, thank you!

ml/main.py
```diff
@@ -0,0 +1,27 @@
...
+    text: List
+
+
+@app.get("/encode/{project}")
+def encode(encode: Encode, project):
+    return {"results": utils.encode(project, encode.text).tolist()}
+
+
+@app.post("/train/{project}")
+def train(project, file: UploadFile = File(...)):
```
AstridKery, line 23: We'll need to add some guard rails around this function for the case where an invalid file is added.

backend/django/core/urls/api.py
```diff
@@ -101,4 +100,5 @@
     re_path(r"^export_database_table/(?P<project_pk>\d+)/$", api.export_database_table),
     re_path(r"^", include(annotate_patterns)),
     re_path(r"^", include(adminpage_patterns)),
+    re_path(r"^download_ml_example/$", api.download_ml_example),
```
AstridKery, line 103: Is this page supposed to be accessible to projects that don't have enough labels to warrant a ML model?

backend/django/core/templates/projects/update/ml.html
```diff
@@ -0,0 +1,52 @@
...
+                {{ form.media.css }}
+                <form action="." method="post" enctype="multipart/form-data" onsubmit="addASpinner()">
+                {% csrf_token %}
+                <h1>ML Settings</h1>
+                <h3>Description</h3>
+                <p>By default, SMART uses a trained machine-learning model to find the best label suggestions for data.</p>
+                <p>If you would like to add weighted text pairs to further train the SMART model for this project, upload a .csv file:</p>
+                <h3>Instructions</h3>
+                <p>Upload a .csv file with weighted text pairs:</p>
+                <p><a href="{% static 'example_ml.csv' %}">An example pairs file can be downloaded from here</a>.</p>
```
AstridKery, line 26: What do the 0.7 and 0.3 in this example file mean? I think we need to add some more information to help SME's to know how to set up these files.

PR #262, opened 2023-08-22
Finish replacing AnnotateCard with DataCard

andykawabata:
### Overview
The original branch `table_query_upgrades` introduced a  new reusable component called `DataCard` and implemented it in the history table replacing the prior `AnnotateCard`. This PR refactors the rest of the tabs in the **Annotate Data** page to use `DataCard` instead of `AnnotateCard`.

### Details
- `DataCard` contains logic to conditionally render its sub-components based on which page it's called from, as well as other factors.
- This component processes the incoming data to a standard format and provides the correct callbacks to its sub-components.
- These callbacks are currently a mix of new react-query hooks (added in original branch) and existing actions.
- Some variable names were changed for clarity/readability.

### Obsolete Components
The following components are made obsolete by this refactor. I've left them in for now for reference.
- `AnnotateCard/index.jsx`
- `DataCard/CardData.jsx`
- `DataCard/index.jsx`
- `History/index.jsx`
- `containers/history_container.jsx`

### Remaining Tasks (Awaiting Confirmation)
- [x] The **History** page is no longer the default when navigating to **Annotate Data**.
- [x] The new history table does not ask "Are you sure" before relabeling.
- [x] The "Requires Adjudication" badge is not updating unless page refreshed.


frontend/src/components/DataCard/ConfirmationModal.jsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { Button, Modal } from "react-bootstrap";
+
+const ConfirmationModal = ({ fn, showModal, setSelectedLabelID }) => {
```
AstridKery, line 4: The only issue I'm seeing here is that if we are in the history table and check the "unlabeled data" check (only will show up for projects without IRR turned on, so not the default projects) the confirmation popup still shows when you're labeling that data even though the data does not have a prior label.

AstridKery:
Thanks so much @andykawabata this looks great! :) I'm not running into any of the errors I noted in the previous merge request.

One problem I did notice is actually not from your changes but a product of how Durk updated the history table. Originally there was an external filter form that someone could fill out and then hit "search" where-upon the table would be loaded with the first "page" of 100 items. There was then a dropdown where they could switch between pages of data. The goal of this was to avoid sending too much to the front end at any time.

Durk does still have the pages, in a subtler form (the pagination at the bottom of the table seems to call for the next page in the same way), but the problem is that he completely dropped the filter form in favor of the in-table filtering, and that only applies to the current page. Our users commonly need to filter by a metadata field like "topic" and label everything in the results, so the filtering needs to search across all available project data.

peridoteagle:
I ran the following QC tests for this PR on a small dataset (n=39) with ~20 labels and large dataset (n~1000) with all CIP codes (10000+):
(1) High priority:  the ML suggested labels seem to only update when I change windows and the come back. See screen recording. The labels do appear correctly in the history and adjudicate tabs.

https://github.com/RTIInternational/SMART/assets/34693652/3f20a894-cecc-41ed-bd6e-20b269ac138e

(2) Medium priority: it'd be helpful to have a label on the box that allows users to select which columns to display in the history table
(3) Low priority: searching Timestamp in the history table by string feels strange, but not sure if there's a good alternative

andykawabata:
> Overall, really nice work! I ran the following QC tests for this PR on a small dataset (n=39) with ~20 labels and large dataset (n~1000) with all CIP codes (10000+): (1) High priority: the ML suggested labels seem to only update when I change windows and the come back. See screen recording. The labels do appear correctly in the history and adjudicate tabs.
> 
>  Screen.Recording.2023-09-19.at.4.47.31.PM.mov 
> (2) Medium priority: it'd be helpful to have a label on the box that allows users to select which columns to display in the history table (3) Low priority: searching Timestamp in the history table by string feels strange, but not sure if there's a good alternative

Thanks @peridoteagle , I just pushed a fix for (1)
For (2) should the label just be "select columns"?
For (3) since we decided to replace the individual column filters with the old filter form for now, this won't be a problem.

AstridKery:
Just tested these updates locally and it looks great! @peridoteagle would you mind taking another pass? If all of your concerns are met, then I think we're good to merge this.

PR #264, opened 2023-11-06
Update to babel 7 to fix fsevents vulnerability

rjruizes:
Currently the frontend has the dependency chain babel 6 > chokidar > fsevents v1 which has a [severe vulnerability](https://github.com/advisories/GHSA-xv2f-5jw4-v95m).

This PR fixes the problem by [updating](https://github.com/babel/babel-upgrade) to babel 7.


AstridKery (approved):
I think this is ready to go

PR #276, opened 2023-11-20
babel-loader upgrade

rjruizes:
Did a `yarn add babel-loader`


AstridKery (approved):
I tested this and I think we're good

PR #285, opened 2023-12-15
fix broken timezone issue and some linting

AstridKery:
This addresses two issues:

1. The download labeled data was broken by one of the recent upgrades. The timezone normalizing was causing an error.
2. A typo was discovered in the history table datetime rendering. %m was being used to render the minutes but %m is month and %M is minute.


rjruizes (approved):
lgtm

PR #293, opened 2024-01-17
Sync requirements.in with requirements.txt, and update dependencies

andykawabata:
**Sync requirements.in with requirements.txt**
- it appeared that `requirements.txt` had been updated independently of `requirements.in` 
- To reconcile this I unpinned some outdated versions `requirements.in` and removed the corresponding entries in `requirements.txt` then re-ran `pip-compile`

**Update Pillow and Sentence-Tranformer**
- Added `Pillow` to `requirements.in` which was compiled to a secure version (10.2.0)
- The above changes allowed `sentence-tranformer` to compile to a secure version (2.2.0)

**Note about pip-compile instructions in README**: The readme specifies running the command `docker-compose run --rm backend pip-compile docker/requirements.in` to compile the requirements. However this method kept giving me a connection error. To run `pip-compile` i had to open a terminal in the backend container and run it. Wondering if this is just a me problem, or if we should update the README to include this.


AstridKery (approved):
I tested this locally and it looks good to me

PR #296, opened 2024-02-02
Upgrade django, jinja2, cryptography, transformers

andykawabata:
**Dependency conflict note:**

Ran into a dependency conflict when upgrading `transformers`.This upgrade caused a bump in `huggingface_hub`  which brought out a known bug in the current version of `sentence-transformers`

There was a patch for this specific bug so I upgraded `sentence-transformers` from 2.2.0 to 2.2.1. I had to pin it there because going to 2.3.x requires upgrading torch - which seems like a big upgrade.

JIRA issue: https://jira.rti.org/browse/EWDCODE-294


AstridKery (approved):
I checked this by creating a new project with many labels and making sure the model still worked properly. As far as I can tell everything is running well. Thanks!

PR #304, opened 2024-03-26
Paginate searchable label dropdown

AstridKery:
This update takes the label dropdown and swaps it out with a searchable version with pagination, so that someone can keep scrolling to populate the dropdown with the next batch of results for their search.

A few other smaller updates meant to help with large label counts:

* for getLabels calls, truncate to first 100 labels. This should be fairly safe as most getLabels calls on the front end revolve around checking if there are < 5 labels and making buttons if there are. 
* Dropping the label guide (but keeping the codebook). As the buttons and dropdown include label descriptions the label guide has been obsolete for a while and would just slow things down.


andykawabata (approved):
This looks good as to me, though I didn't get a chance to test anything with the codebook code.

I tried for a bit to implement react-query hooks for the search_labels query, just for consistency, but I think just using fetch directly as you did is a lot cleaner and more readable.

PR #313, opened 2024-04-22
Fix "Add project to group" Bug

andykawabata:
Replaced deprecated Django template syntax


AstridKery (approved):
This fixed the issue. Thanks so much!

PR #316, opened 2024-04-26
Add some tweaks to the project creation code

AstridKery:
* remove full file limit (allow users to upload files with no unlabeled data)
* when someone uploads a file with labels not defined in step 2, just print the extra labels instead of both full label sets (this works much better when there are many labels)

### DIFF
--- a/backend/django/core/templates/projects/create/create_wizard_labels.html
+++ b/backend/django/core/templates/projects/create/create_wizard_labels.html
@@ -31,19 +31,22 @@
           <p>In the <strong>Labels</strong> section, we will create categories for labeling. These labeled observations will be used to train a classification model that predicts which of these categories a new observation is most likely to be.</p>
           <h3>Instructions</h3>
           <p>Please fill-in below the names of the categories you are interested in predicting.  If you have more than two labels, use the <code>add label</code> button to add more rows to the form. If you decide that you want to remove a label after adding it, use the <code>remove label</code> button to remove the label name.</p>
+          <p>You may also upload a .csv file containing labels and their descriptions (label and description columns required).</p>
+          
           <p><i>Labeling Notes:</i></p>
           <ul class="list-group">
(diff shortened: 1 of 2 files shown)


AstridKery:
Things to look for in this merge request

* when someone is looking to upload data to the project, SMART should no longer throw an error if the file only has labeled data. Reviewers should make sure that uploading a full file when creating the project doesn't lead to any unexpected behavior downstream in the project (i.e. when someone goes to the annotate page, when someone uploads unlabeled data), especially in the cases where the modeling or active learning features are turned on.
* try uploading a label file with a new label not defined in step 2 and see if the error printed to the user makes sense
* look at the label upload directions and see if they make sense

backend/django/core/templates/projects/create/create_wizard_labels.html
```diff
@@ -31,19 +31,22 @@ <h3>Description</h3>
...
+          <p>You may also upload a .csv file containing labels and their descriptions (label and description columns required).</p>
+          
           <p><i>Labeling Notes:</i></p>
           <ul class="list-group">
             <li class="list-group-item">SMART <strong>requires at least two category labels</strong> and the labels must be <strong>unique</strong>.</li>
             <li class="list-group-item">If you plan on uploading a data file that contains labels, the label categories in the file must match those provided on this page.</li>
-            <li class="list-group-item">You may add up to 10,000 labels to each project.</li>
+            <li class="list-group-item">SMART has been tested with up to 50,000 labels.</li>
             <li class="list-group-item">You cannot update the labels for a project after the project is created.</li>
+            <li class="list-group-item">Any labels currently in the table will be overwritten by the file data if you upload a .csv file.</li>
```
andykawabata: Saying "table below" instead of just "table" might make this more clear

andykawabata (approved):
This looks good as far as I can tell.
- I uploaded a fully labelled dataset and could not break anything while clicking through annotation pages
- The error message when uploading data with bad labels was clear. Worked both on the initial data upload and when updating data.
- The label upload instruction seemed mostly clear. I did leave one comment with a suggestion.

PR #328, opened 2024-06-05
Handle bug where the same card was appearing multiple times

AstridKery:
This MR addresses a bug which was occurring whenever someone either:
* finished the batch of cards handed out
* double clicked on a card

Either action would cause SMART to call `fetchCards` multiple times which would lead to the same card deck being pushed two to three times to the card state on the front end. 

I looked around for the reason this was happening. Even if you comment out all calls to fetchCards that run after actions and only leave the one in the CardContainer constructor the double call still occurs (which suggests to me that the constructor call is the source. @andykawabata any thoughts on where this call should go so that fetchCards runs when the component is mounted?). 

This MR is a workaround that makes the pushCard action check if a card already exists in the deck before pushing. It also adds more error handling for double-clicking since sometimes the first check doesn't catch it.

### DIFF
--- a/frontend/src/reducers/card.js
+++ b/frontend/src/reducers/card.js
@@ -30,12 +30,13 @@ const card = handleActions({
         return update(state, { cards: { $splice: [[0, 1]] } } );
     },
     [PUSH_CARD]: (state, action) => {
-        // Set the start time of the new top card to the current time
+        // only push the card if it's not already in the stack - failsafe
         for (let i = 0; i < state.cards.length; i++) {
             if (state.cards[i].id == action.payload.id) {
                 return state;
             }
         }
+        // Set the start time of the new top card to the current time
         if (state.cards.length > 0) {
             state.cards[0]['start_time'] = moment();
         }
--- a/frontend/src/reducers/card.js
+++ b/frontend/src/reducers/card.js
@@ -11,15 +11,9 @@ const initialState = {
 
 const card = handleActions({
     [POP_CARD]: (state, action) => {
-        // if the card isn't in the deck don't pop it off
+        // if the card isn't the first item in the deck don't pop it off
         // This handles double-clicking of Skip
-        let found_card = false;
-        for (let i = 0; i < state.cards.length; i++) {
-            if (state.cards[i].text.pk == action.payload) {
-                found_card = true;
-            }
-        }
-        if (! found_card) {
+        if (state.cards[0].text.pk != action.payload) {
             return state;
         }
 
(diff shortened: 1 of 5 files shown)


andykawabata:
@AstridKery 
> Even if you comment out all calls to fetchCards that run after actions and only leave the one in the CardContainer constructor the double call still occurs 

On `dev` branch I tried commenting this line, and it seemed to fix the issue
https://github.com/RTIInternational/SMART/blob/0824189ce56b46bc0face38f61c34c7c849bc6f3/frontend/src/actions/card.js#L79

Though I have only been testing for the case of "finished the batch of cards handed out" - not sure what you mean by "double click on a card"

AstridKery:
> Though I have only been testing for the case of "finished the batch of cards handed out" - not sure what you mean by "double click on a card"

@andykawabata I'm talking about if you click a label button twice in quick succession when labeling a card. It's a common thing for users to accidentally do. It sends two calls of annotate_card to the backend which returns an error and forces the deck to refresh (the error is also meant to help in cases where someone's cards were un-assigned by an admin). 

For the line you indicated, that's one of the calls that run after actions that I was talking about. There's one of those for skipping or adjudicating cards as well. I tried commenting out all of those but when I added `console.log(props.cards)` to https://github.com/RTIInternational/SMART/blob/0824189ce56b46bc0face38f61c34c7c849bc6f3/frontend/src/containers/card_container.jsx#L12 it still seemed like the deck was being double-filled at the end of each batch.... I'll check again real quick.

andykawabata:
@AstridKery 
Oh I see - I was only testing with a project with 3 labels. It looks like it's more of a issue with projects with label select/dropdown. Will continue looking it to it.

frontend/src/reducers/card.js
```diff
@@ -10,14 +10,32 @@ const initialState = {
 };
 
 const card = handleActions({
-    [POP_CARD]: (state) => {
+    [POP_CARD]: (state, action) => {
```
andykawabata, line 13: It looks like we are now are selecting the card by ID, but still removing the card at index 0 no matter what. Fixed in this PR: https://github.com/RTIInternational/SMART/pull/330

AstridKery: I suppose? I'm passing card so that it would only pop the first item off if the card being popped existed. I'll look at your MR.

andykawabata: Ah I see, you can ignore that PR then.

andykawabata: Maybe just add a note that if it does find the card, we are confident that it will be at index 0

AstridKery: I just pushed a change that only checks if the card at 0 is the right one rather than searching the whole deck

andykawabata (approved):

PR #340, opened 2024-06-24
Allow sorting across all data in history table

andykawabata:
Updated the `get_label_history` view to sort across all data. Previously you could only sort the data in each page.

I was able to optimize the function for sorting by the "Data" column, however sorting by the other columns is significantly slower.

### DIFF
--- a/backend/django/core/views/api_annotate.py
+++ b/backend/django/core/views/api_annotate.py
@@ -1030,8 +1030,7 @@ def get_label_history(request, project_pk):
         current_page = 1
     page = int(current_page) - 1
 
-    page_size = 100
-    all_data = Data.objects.filter(pk__in=total_data_list).order_by("text")
+    all_data = Data.objects.filter(pk__in=total_data_list)
     metadata_objects = MetaDataField.objects.filter(project=project)
 
     # filter the results by the search terms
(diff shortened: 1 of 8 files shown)


backend/django/core/views/api_annotate.py
```diff
@@ -1129,12 +1138,13 @@ def get_label_history(request, project_pk):
...
-    # now add back on the metadata fields
-    results = data_df.fillna("").to_dict(orient="records")
-    for i in range(len(results)):
-        results[i]["metadata"] = all_metadata[i]
-        results[i]["formattedMetadata"] = all_metadata_formatted[i]
-        results[i]["metadataIDs"] = page_data_metadata_ids[i]
+    if pre_sorted:
+        page_data = data_df
+    else:
+        df_sorted = data_df.sort_values(by=[sort_by, "data"], key=lambda col: col.str.lower(), ascending=(not reverse))
```
AstridKery: Looking into this some more, we have a couple of reasons why this is much slower when sorting by other fields

1. We end up pulling metadata, Labels, IRRLogs for all data instead of the first page of it because the pagination call is moved to the very end.
2. We are using the pandas's sorting method instead of django's

I think the first one is the main culprit though. In a large project with no filters applied, that's going to be a lot of data to pull info for.

One way to speed this up would be to look for code which isn't needed for the new sorting function and move it to after the iloc line. For example: currently you cannot sort by metadata, so you can probably move the metadata calls to only run on the one-page dataset.

AstridKery (approved):
Looks good!

PR #342, opened 2024-06-27
Display each user's labels when adjudicating IRR mismatches

andykawabata:
- created a new endpoint to retrieve the irr log entries that also appear in the admin queue
- the adjudication tab pulls from this endpoint and displays a table each user's labels for IRR disagreements


AstridKery (approved):
This looks perfect! Great job

PR #345, opened 2024-07-02
Deploy recent updates to production 

AstridKery:
This branch contains the following updates. I'd like these to be given a thorough review by someone other than Andy and I before we deploy:

* remove pagination on project list page
* add sorting to history table for fields not including medatada
* add setting to project which allows people with "coder" permissions to see all labeled data (this setting can be changed in the update form)
* allow for un-verification of data and lock label changes in the history table when something is verified


michael-long88 (approved):
Everything looks good! Tested out with a few different users and all of the behavior seems to match the code. Thanks for this!

PR #346, opened 2024-07-03
add IRR log download option -- MERGE AFTER PR #342

andykawabata:
Add ability to download IRR logs - button is on project details page. Only appears if the project uses IRR.

This branch is based off branch `irr-label-table` which has  PR (#342)


AstridKery (approved):
Looks good

PR #352, opened 2024-07-23
Add type column to history table

andykawabata:
Implemented logic for type column in history table. The possible types are
- Labeled
- Unlabeled
- IRR Incomplete
- IRR Pending
- IRR Finalized
- IRR Historic

Currently the "Type" column is text - I believe we'll want to change that to icons before merging.

### DIFF
--- a/backend/django/core/views/api_annotate.py
+++ b/backend/django/core/views/api_annotate.py
@@ -999,37 +999,54 @@ def get_label_history(request, project_pk):
     project = Project.objects.get(pk=project_pk)
 
     labels = Label.objects.all().filter(project=project)
-    # irr data gets set to not IRR once it's finalized
-    finalized_irr_data = IRRLog.objects.filter(data__irr_ind=False).values_list(
-        "data__pk", flat=True
-    )
+
+    permission_filter = Q()
+    if project_extras.proj_permission_level(project, profile) < 2:
+        permission_filter &= Q(profile=profile)
+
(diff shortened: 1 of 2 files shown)


backend/django/core/views/api_annotate.py
```diff
@@ -996,37 +996,54 @@ def get_label_history(request, project_pk):
...
-    # irr data gets set to not IRR once it's finalized
-    finalized_irr_data = IRRLog.objects.filter(data__irr_ind=False).values_list(
-        "data__pk", flat=True
-    )
+
+    permission_filter = Q()
+    if project_extras.proj_permission_level(project, profile) < 2:
+        permission_filter &= Q(profile=profile)
+
+    incomplete_irr_data =  DataLabel.objects.filter(profile=profile, data__project=project_pk, data__irr_ind=True).values_list("data__pk", flat=True)
```
AstridKery: One issue here. If someone chooses to "adjudicate" an IRR item it actually doesn't show up in the DataLabel table but instead in the IRRLog (this is because the DataLabel table doesn't allow for null labels but the IRRLog does). 

Right now, those items are showing up for admin in their history tables as "IRR Pending" even if they were not the one to mark it for adjudication. This isn't as dangerous as it would be if they were able to see other coders incomplete labels, but it might confuse people. 

I would recommend using the admin queue as a method for distinguishing incomplete versus pending IRR. When IRR data get processed and a disagreement is sent to the admin for resolution the data is added to the admin queue.

andykawabata: Just pushed a fix for this

AstridKery (approved):
Looks good to me. Thanks!

PR #353, opened 2024-07-24
Add metadata to labels and add project category which filters suggestions

AstridKery:
This branch scraps the current label upload/update page and instead changes it to a label file upload only. In addition it adds label metadata fields in the same vein as data metadata fields. If data and label have a metadata field in common, that field is chosen as the "Category" by default and this impacts which labels are suggested for which data. The category can be changed or removed in the advanced details update page.


AstridKery:
One note: I need to fix the carrot opening/closing on the update page. Right now the top one affects both sections

andykawabata (approved):

PR #356, opened 2024-07-30
Fix bug causing projects with > 100 labels to break

andykawabata:
Previously the IRR tables were relying on the `get_labels` endpoint to retrieve label names to join to the ids. With this update, we include the label names in the IRR endpoint so no need to call `get_labels`.


AstridKery (approved):
Looks good to me. Thanks!

PR #357, opened 2024-07-30
Temporarily disable Incomplete IRR labeling ability causing unexpected behavior

andykawabata:
For IRR Incomplete data that has been adjudicated - it has no label but still shows up in history table and allows you to change label. Changing the label causes unexpected behavior. As a temporary fix I have disable editing for IRR Incomplete items with no label.

Also pressing "adjudicated" on an IRR incomplete item causes unexpected behavior. As a temporary fix I disable the adjudicate button for IRR Incomplete data.


AstridKery (approved):
Looks good, thanks!

PR #359, opened 2024-08-15
Refactor skew search feature to fix bug

andykawabata:
Previously, on the skew page if you did a search and then labeled something, the data in the table would not repopulate like it would if you hadn't done a search first.

This update changes how the filter is applied so that it happens within the same api call that retrieves the data. This way we do not have to perform 2 separate actions for pulling unlabeled data with filter vs pulling all unlabeled data.


AstridKery (approved):
Looks good. I think in the future we'll want to add pagination to the skew page filtering but fixing the bug is priority for now.

PR #361, opened 2024-09-03
Prevent history table from truncating rows at 100

andykawabata:
With the new IRR logic, sometimes over 100 rows get returned for each history page. Previously, the History Table frontend was configured to cap rows at 100. This PR updates the config allowing it to render all rows for page.


AstridKery (approved):
Looks good. Thanks!

PR #362, opened 2024-09-04
add a whole bunch of category fixes

AstridKery:
This MR pulls together a whole bunch of little updates, mostly related to the recent category update:

* moves the label serializer code to only run in get_labels if labels < 100
* makes the category case-insensitive, so that the metadata fields with different capitalization in their name can be matched
* adds loading spinners to the label upload page and the project creation page to avoid confusion
* BUG FIX: closes a loophole in the annotate_data function so that if someone just labeled the item in the history table it won't make a second label
* slightly modifies the codebook rendering to not run a bunch of the slower code when there is no codebook
* adds an "all" category to the label category dropdown so that someone can search all labels if needed
* fixes the null description/metadata rendering so that they just aren't shown instead of having "nan" everywhere

### DIFF
--- a/backend/django/core/forms.py
+++ b/backend/django/core/forms.py
@@ -155,9 +155,11 @@ class ProjectUpdateAdvancedForm(forms.ModelForm):
         else:
             start_val = "None"
         data_metadata = project.metadatafields.values_list("field_name", flat=True)
+        data_metadata = [m.lower().capitalize() for m in data_metadata]
         label_metadata = project.labelmetadatafields.values_list(
             "field_name", flat=True
         )
+        label_metadata = [lm.lower().capitalize() for lm in label_metadata]
         options = list(set(data_metadata) & set(label_metadata)) + ["None"]
 
         percentage_irr = kwargs.pop("percentage_irr")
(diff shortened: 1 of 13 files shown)


andykawabata (commented):
This looks good to me. The only thing I did that kind of broke it was
1) Create a project with matching data and label metadata (category column = State)
2) Go to the update label descriptions page
3) Re-upload the label file but the category column name is changed to sTatE
Results in screenshot below. But I feel like it might be safe to assume that no one would do this?
Screenshot 2024-09-05 at 12 02 35 PM

AstridKery:
> This looks good to me. The only thing I did that kind of broke it was
> 
> 1. Create a project with matching data and label metadata (category column = State)
> 2. Go to the update label descriptions page
> 3. Re-upload the label file but the category column name is changed to sTatE
>    Results in screenshot below. But I feel like it might be safe to assume that no one would do this?
>    Screenshot 2024-09-05 at 12 02 35 PM

Thanks! I had been toying with adding case-insensitive checks to the file reads. This confirms that it would be a good idea. I didn't even think of someone just changing a field name, but that's possible. I've added that in now.

andykawabata (approved):

PR #367, opened 2024-11-13
fix metadata editing bug

AstridKery:
This fixes two bugs:

* null metadata values being rendered as "None" and then throwing an error if edited
* metadata field names and values now won't have weird behavior if they contain colons ":"

This mostly meant fixing the metadata serializer to return the objects instead of a <key>:<vlue> string and then updating all of the places that use that serializer to account for that.


andykawabata (approved):
This PR solves both issues as far as I can tell. I also tried merging it with the `fix_category_sorting_bug` branch and no issues.

PR #370, opened 2024-11-25
Handle IRR assignments

AstridKery:
Right now if the number of IRR labels needed for each data is 2 and three coders are online, the item will get passed to all three of them. This update limits that so that an item isn't passed out to that third coder if the number of coders required for each data item is 2.

It also adds back in a message to display when no cards are returned to the front end as requested by users.

### DIFF
--- a/backend/django/core/utils/utils_annotate.py
+++ b/backend/django/core/utils/utils_annotate.py
@@ -96,15 +96,19 @@ def get_assignments(profile, project, num_assignments):
         for i in range(num_assignments):
             # if there is IRR, with some probability get an IRR item
             rand_choice = randrange(0, 101)
+
+            assigned_datum = None
             if project.percentage_irr > 0 and rand_choice <= project.percentage_irr:
                 assigned_datum = assign_datum(profile, project, type="irr")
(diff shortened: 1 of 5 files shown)


andykawabata (approved):
In addition to running the pytests, I manually tested this pull request by simulating an IRR project with three coders. I tried this with 50% and 100% IRR projects, each with a two-coder limit. I was mainly testing for three things:

1. **No datum should be assigned to more than two coders at a time under any circumstances.**
2. **The new restriction on data assignment doesn't prevent data that should be assigned from getting assigned.**
3. **IRR still works as expected.**

I had all three coders enter the annotation page before assigning any labels and then tried various combinations of labeling, skipping, and adjudicating while cross-checking the database tables to make sure everything was in the expected state.

As far as I could tell, everything is working as intended, as I was unable to find any circumstance where a datum could be labeled by more than two coders.

**Note on Pytests**

I did notice a limitation in tests `irr_test.py` where after two agreeing labels, we should be able to assert that `datum.irr_ind is False`. However, it does not actually get changed to `False` since we are only calling `label_data()` and not everything else that gets called when actually annotating.

